PR #3, opened 2021-06-28
Eval cb

olliethomas:
I've not added a test, but I think that this is basically the same as in ED

### DIFF
--- a/bolts/callbacks/posthoc_eval.py
+++ b/bolts/callbacks/posthoc_eval.py
@@ -1,0 +1,75 @@
+from collections import Sequence
+
+import pytorch_lightning as pl
+import torch
+from kit import gcopy
+from pytorch_lightning.utilities.types import STEP_OUTPUT
+from torch.optim import Optimizer
+from torch.utils.data import DataLoader
+
+
+class PostHocEval(pl.Callback):
+    """
+    Trains and evaluates an auxillary classifier model.
+    Your model should have:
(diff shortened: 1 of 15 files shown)


bolts/callbacks/posthoc_eval.py
```diff
@@ -0,0 +1,75 @@
...
+    Your model should have:
+        - ``self.classifier``
+        - ``self.datamodule``
+        - ``self.clf_epochs``
+        - ``self.trainer''
+    """
+
+    def __init__(self) -> None:
+        super().__init__()
+        self.optimizer: Optimizer
```
MylesBartlett, line 23: This is (rightfully) unused

MylesBartlett (approved):

bolts/callbacks/posthoc_eval.py
olliethomas: yeah... should the classifier be an nn.Module, or a pl.LightningModule? I guess the latter - makes life easier all round

MylesBartlett: Keep it as it is for now. While I don't much like the approach because of the assumptions it makes, I don't have in mind a good alternative at the moment.

olliethomas: yeah... Looking in the bolts repo they follow this setup of just palming off the responsibility. Maybe this is the best way 🤷‍♂️

PR #13, opened 2021-07-02
Add NICO dataset with basic data-loading/downloading functionality (no biasing schemes yet).

MylesBartlett:


bolts/data/datasets/utils.py
```diff
@@ -71,7 +73,7 @@ class BinarySample(NamedTuple):
 
 class TernarySample(NamedTuple):
     x: Tensor | np.ndarray | Image.Image
-    s: Tensor | float
+    z: Tensor | float
```
thomkeh, line 76: ?

bolts/data/datasets/vision/nico.py
```diff
@@ -0,0 +1,152 @@
...
+        return metadata
+
+    def __len__(self) -> int:
+        return len(self.x)
+
+    def __getitem__(self, index: int) -> TernarySample:
+        image = load_image(self._base_dir / self.x[index], backend=self._il_backend)
+        image = apply_image_transform(image=image, transform=self.transform)
+        target = self.y[index]
+        return TernarySample(x=image, s=self.z[index], y=target)
```
thomkeh, line 152: It still says `s=` here.

bolts/data/datasets/utils.py
MylesBartlett: ?

thomkeh: Why would you give s such a confusing name?

MylesBartlett: For the very reason that it's not s - it's more general than that. z is hardly an arbitrary choice to go with x and y and indeed it's fairly common in the literature to use it to denote some side information.

olliethomas: this will cause problems for `bolts.fair.models` which assume `batch.x`, `batch.s` and `batch.y`

PR #18, opened 2021-07-06
give option to not copy every callback to the inner trainer

olliethomas:


olliethomas:
small patch - not looking for review. Just wanted to trigger the CI

PR #22, opened 2021-07-07
add parsable to datamodules

olliethomas:
requesting to see if you agree, that only the datamodules are  what you would want to parse with configen


thomkeh:
minimial `palkit` version needs to be changed in `pyproject.toml`, right?

olliethomas:
> minimial `palkit` version needs to be changed in `pyproject.toml`, right?

`palkit = "*"` is the current minimal version.

thomkeh (approved):

PR #23, opened 2021-07-07
Add NICO data-module.

MylesBartlett:


bolts/data/datamodules/vision/nico.py
```diff
@@ -0,0 +1,89 @@
+"""CelebA DataModule."""
```
thomkeh, line 1:
```suggestion
"""NICO DataModule."""
```

bolts/data/datasets/vision/celeba.py
```diff
@@ -45,8 +48,8 @@ def __init__(
 
         dataset, self._img_dir = em.celeba(
             download_dir=root,
-            label=superclass,
-            sens_attr=subclass,
+            label=superclass,  # type: ignore
+            sens_attr=subclass,  # type: ignore
```
thomkeh, line 52: Could use a TypeGuard here, if you wanted to be fancy

bolts/fair/data/datamodules/base.py
```diff
@@ -1,17 +1,20 @@
...
 import logging
-from typing import List, Optional, Sequence, Union
+from typing import Optional, Sequence, Union
 
 from kit import implements
 from kit.torch import StratifiedSampler
 import pytorch_lightning as pl
 from torch.utils.data import DataLoader, Dataset, Sampler
 
+from bolts.data.datamodules import BaseDataModule as _BaseDataModule
```
thomkeh, line 10: Is this right? It looks like a circular import.

bolts/fair/data/datamodules/vision/celeba.py
```diff
@@ -4,26 +4,28 @@
...
-from .base import VisionBaseDataModule
+from .base import BaseVisionDataModule
 
 __all__ = ["CelebaDataModule"]
 
 
-class CelebaDataModule(VisionBaseDataModule):
-    """CelebA Dataset."""
+class CelebaDataModule(BaseVisionDataModule):
+    """Fairness-oriented CelebA data-module."""
```
thomkeh, line 22: Eventually this datamodule should use the CelebA dataset class I added, right?

bolts/data/datamodules/base.py
```diff
@@ -0,0 +1,73 @@
...
+
+    _train_data: Dataset
+    _val_data: Dataset
+    _test_data: Dataset
+
+    def __init__(
+        self,
+        batch_size: int,
+        val_prop: Union[float, int],
+        test_prop: Union[float, int],
```
olliethomas, line 25: these can only be floats though? as `train_prop = 1 - val_prop - test_prop`

bolts/data/datamodules/common.py
```diff
@@ -0,0 +1,7 @@
+from __future__ import annotations
+
+from typing_extensions import Literal
+
+__all__ = ["Stage"]
+
+Stage = Literal["fit", "validate", "test"]
```
olliethomas, line 7: we use different terminology elsewhere (train, val, test) - we should consolidate

MylesBartlett: We do, and incorrectly. I checked the lightning docs last night - what I have here is the correct set of options for Stage.

bolts/fair/data/datamodules/vision/celeba.py
MylesBartlett: Yes but it's beyond the scope of this PR

bolts/fair/data/datamodules/base.py
MylesBartlett: it's not a circular import - what makes you think so? The fair modules depend on the non-fair ones, never the converse.

thomkeh: oh, I missed the `.fair` part, sorry

thomkeh: still, it might not be the best idea to have two identically-named classes in the code base

PR #24, opened 2021-07-08
Types

olliethomas:


bolts/common.py
```diff
@@ -1,7 +1,31 @@
 from __future__ import annotations
+import collections
+from typing import Generic
 
-from typing_extensions import Literal
+from typing_extensions import ClassVar, Literal
 
 __all__ = ["Stage"]
 
+from typing_inspect import typingGenericAlias
```
thomkeh, line 9: wait, is `typing_inspect` a dependency? then you can just import `get_args` from there. I copied it from there.

bolts/fair/models/gpd.py
```diff
@@ -1,19 +1,21 @@
...
-from torch.optim.lr_scheduler import CosineAnnealingWarmRestarts, _LRScheduler
+from torch.optim.lr_scheduler import CosineAnnealingWarmRestarts
 import torchmetrics
 from torchmetrics import MetricCollection
 
-from bolts.common import Stage
+from bolts.common import Stage, get_args
 from bolts.fair.data import DataBatch
 from bolts.fair.losses import CrossEntropy
+from bolts.fair.models import LRScheduler, SchedInterval
```
thomkeh, line 18: this risks circular import otherwise
```suggestion
from bolts.fair.models.utils import LRScheduler, SchedInterval
```

thomkeh (approved):
approve except for the question of whether `typing_inspect` will be used

bolts/common.py
olliethomas: `ethicml > typed-argument-parser > typing-inspect`

thomkeh: I see. Maybe we should get rid of that dependency on typed-argument-parser.

In any case, I don't think it makes sense to do a mix of copying from typing-inspect and using typing-inspect as a library.

olliethomas: I'll use `typing-inspect` until we get rid of it from ethicml.

PR #28, opened 2021-07-09
fair mixup model

olliethomas:
This is actuallly only the demographic parity mdel. There is an Equal Opprotunity one too.

- wip
- add fairn mixup model


olliethomas:
Code mostly comes from [here](https://github.com/chingyaoc/fair-mixup/blob/master/celeba/main_dp.py)

bolts/fair/models/fair_mixup.py
```diff
@@ -0,0 +1,286 @@
...
+                f"train/loss": loss_sup.item(),
+                f"train/acc": self.train_acc,
+            }
+        )
+
+        ops = logits.sum()
+
+        # Smoothness Regularization
+        gradx = torch.autograd.grad(ops, inputs, create_graph=True)[0].view(inputs.shape[0], -1)
+        x_d = (x_b - x_a).view(inputs.shape[0], -1)
```
MylesBartlett, line 90:
```suggestion
        x_d = (x_b - x_a).view(inputs.size(0), -1)
```

bolts/fair/models/fair_mixup.py
```diff
@@ -0,0 +1,286 @@
...
+        grad_inn = (gradx * x_d).sum(1).view(-1)
```
MylesBartlett, line 91:
```suggestion
        grad_inn = (gradx * x_d).sum(1).flatten()
```

bolts/fair/models/fair_mixup.py
```diff
@@ -0,0 +1,286 @@
...
+            {
+                f"{stage}/loss": loss.item(),
+                f"{stage}/{self.target}_acc": _acc,
+            }
+        )
+
+        return {
+            "y": batch.y.view(-1),
+            "s": batch.s.view(-1),
+            "preds": logits.softmax(-1)[:, 1],
```
MylesBartlett, line 120: Why do you index the first element of the first dimension here instead of just computing the argmax?

bolts/fair/models/fair_mixup.py
```diff
@@ -0,0 +1,286 @@
...
+    ) -> dict[str, float]:
+        all_y = torch.cat([_r["y"] for _r in results], 0)
+        all_s = torch.cat([_r["s"] for _r in results], 0)
+        all_preds = torch.cat([_r["preds"] for _r in results], 0)
+
+        mean_preds = all_preds.mean(-1)
+        mean_preds_s0 = all_preds[all_s == 0].mean(-1)
+        mean_preds_s1 = all_preds[all_s == 1].mean(-1)
+
+        dt = em.DataTuple(
```
MylesBartlett, line 134: We have this same code repeated in several models; can share the code somehow?

bolts/fair/models/fair_mixup.py
```diff
@@ -0,0 +1,286 @@
...
+        torch.randperm(batch_s1.size()[0]).to(device)
+
+        xal = []
+        xbl = []
+        sal = []
+        sbl = []
+        yal = []
+        ybl = []
+
+        for x_a, s_a, y_a in zip(batch.x, batch.s, batch.y):
```
MylesBartlett, line 235: This seems very inefficient.

bolts/fair/models/fair_mixup.py
```diff
@@ -0,0 +1,286 @@
...
+            else torch.tensor([mix_lambda]).to(device)
+        )
+
+        batch_s0 = batch.x[batch.s.view(-1) == 0].to(device)
+        batch_s1 = batch.x[batch.s.view(-1) == 1].to(device)
+
+        batch_y_s0 = batch.y[batch.s.view(-1) == 0].to(device)
+        batch_y_s1 = batch.y[batch.s.view(-1) == 1].to(device)
+
+        torch.randperm(batch_s0.size()[0]).to(device)
```
MylesBartlett, line 225: This code doesn't do anything

bolts/fair/models/fair_mixup.py
olliethomas: because they want to report the mean prediction. I don't think we really care about this though

bolts/fair/models/fair_mixup.py
```diff
@@ -0,0 +1,286 @@
...
+                xbl.append(x_b)
+                sbl.append(torch.ones_like(s_a).unsqueeze(-1).float())
+                y_b = batch_y_s1[idx].float()
+                ybl.append(y_b)
+
+            elif s_a == 1:
+                xal.append(x_a)
+                sal.append(s_a.unsqueeze(-1).float())
+                yal.append(y_a.unsqueeze(-1).float())
+                idx = torch.randint(batch_s0.size()[0], (1,))
```
MylesBartlett, line 251:
```suggestion
                idx = torch.randint(batch_s0.size(0), (1,))
```

bolts/fair/models/fair_mixup.py
```diff
@@ -0,0 +1,286 @@
...
+        sbl = []
+        yal = []
+        ybl = []
+
+        for x_a, s_a, y_a in zip(batch.x, batch.s, batch.y):
+            if s_a == 0:
+                xal.append(x_a)
+                sal.append(s_a.unsqueeze(-1).float())
+                yal.append(y_a.unsqueeze(-1).float())
+                idx = torch.randint(batch_s1.shape[0], (1,))
```
MylesBartlett, line 240:
```suggestion
                idx = torch.randint(batch_s1.size(0), (1,))
```

bolts/fair/models/fair_mixup.py
```diff
@@ -0,0 +1,286 @@
...
+                x_b = batch_s1[idx, :].squeeze(0)
+                xbl.append(x_b)
+                sbl.append(torch.ones_like(s_a).unsqueeze(-1).float())
+                y_b = batch_y_s1[idx].float()
+                ybl.append(y_b)
+
+            elif s_a == 1:
```
MylesBartlett, line 244: This is the same as the s == 0 branch except for one variable - if you're not going to vectorize the code, you should at least write a local func

olliethomas: how would I go about vectorising it? I need a random pairing

bolts/fair/models/fair_mixup.py
```diff
@@ -0,0 +1,307 @@
...
+
+
+class FairMixup(pl.LightningModule):
+    def __init__(
+        self,
+        enc: nn.Module,
+        clf: nn.Module,
+        lr: float,
+        weight_decay: float,
+        fairness: str,
```
thomkeh, line 31: do you want to pass in a string for backwards compatibility?

bolts/fair/models/fair_mixup.py
```diff
@@ -0,0 +1,307 @@
...
+
+        mix_stats = {
+            "batch_stats/S0=sS0": sum(a == b for a, b in zip(s_a, s_b)) / batch.s.size(0),
+            "batch_stats/S0!=sS0": sum(a != b for a, b in zip(s_a, s_b)) / batch.s.size(0),
+            "batch_stats/all_s0": sum(a + b == 0 for a, b in zip(s_a, s_b)) / batch.s.size(0),
+            "batch_stats/all_s1": sum(a + b == 2 for a, b in zip(s_a, s_b)) / batch.s.size(0),
+        }
+
+        mixed_x = lam * x_a + (1 - lam) * x_b
+        return mixed_x.requires_grad_(True), x_a, x_b, s_a, s_b, y_a, y_b, lam, mix_stats
```
thomkeh, line 296: a namedtuple might be nice with so many returns

bolts/fair/models/fair_mixup.py
```diff
@@ -0,0 +1,307 @@
...
+        yal = []
+        ybl = []
+
+        for x_a, s_a, y_a in zip(batch.x, batch.s, batch.y):
+            xal.append(x_a)
+            sal.append(s_a.unsqueeze(-1).float())
+            yal.append(y_a.unsqueeze(-1).float())
+            if fairness is FairnessType.DP:
+                idx = torch.randint(eval(f"batch_x_s{1-int(s_a)}").size(0), (1,))
+                x_b = eval(f"batch_x_s{1-int(s_a)}")[idx, :].squeeze(0)
```
thomkeh, line 254: `eval()` is very slow and also the opposite of safe. I think you should just use a dictionary `batch_x_s[0]` etc. this would also allow using loops for the masking code above.

thomkeh (changes requested):
most unhappy with the `eval()`

olliethomas:
@thomkeh re-requested as made changes requested

thomkeh (approved):
Looks mostly good now (except for the potentially slower loop)

PR #34, opened 2021-07-27
add s/y dim to dm

olliethomas:


MylesBartlett:
Can you make these properties that yield an error in the same way input_size does when accessed before the DM has been set up?

PR #40, opened 2021-07-30
Add a base model class to reduce code-duplication.

MylesBartlett:


olliethomas:
this base module is kind of heading towards being a [pl-flash Task](https://github.com/PyTorchLightning/lightning-flash/blob/master/flash/core/model.py#L103)

olliethomas:
The way I see it, there are 2 approaches. 

1. Each model is standalone, a bit like the pl-bolts repo where someone can come along and look at the implementation in one of these files as the canonical version of method, without having to look at the base class for some methods.
2. Treat these models as a collection to be applied, more akin to the pl-flash library, where these are still the methods, but they're meant to be used as part of a coherent library more than as a collection of individual models.

Tbh, I don't know the best approach

PR #43, opened 2021-08-02
Convert dataclasses to dictionaries in pb_default_collate instead of using astuple to avoid deep-copying.

MylesBartlett:


bolts/data/datasets/utils.py
```diff
@@ -204,6 +204,17 @@ def compute_instance_weights(dataset: Dataset) -> Tensor:
...
 
+def pb_astuple(dataclass: object) -> tuple[Any, ...]:
+    """
+    dataclasses.astuple() but without deep-copying and recursion.
+    This function has the more limited intended functionality of just extracting fields
+    (in order) into an tuple.
+    """
+    if not is_dataclass(dataclass):
+        raise TypeError("pb_astuple() should be called on dataclass instances")
+    return tuple(getattr(dataclass, field.name) for field in fields(dataclass))
```
olliethomas, line 216: is there any way of ensuring that `for field in fields(dataclass)` retains the right order (x, y, s, iw)

PR #48, opened 2021-08-11
add enum str method to stop calling .value everywhere

olliethomas:


bolts/structures.py
```diff
@@ -17,6 +17,9 @@ class Stage(Enum):
     validate = "validate"
     test = "test"
 
+    def __str__(self):
```
MylesBartlett, line 20:
```suggestion
    def __str__(self) - > str:
```

MylesBartlett (approved):

PR #50, opened 2021-08-12
Add tabular transforms and base tabular dataset.

MylesBartlett:


bolts/transforms/tabular.py
```diff
@@ -0,0 +1,166 @@
+from __future__ import annotations
```
olliethomas, line 1: this dir needs an __init__.py

bolts/transforms/tabular.py
```diff
@@ -0,0 +1,178 @@
...
+        data /= self.new_range + self._EPS
+        data *= self.orig_range
+        data += self.orig_min
+
+    @implements(TabularTransform)
+    def _transform(self, data: Tensor) -> None:
+        data -= self.orig_min
+        data /= self.orig_range.clamp_min(self._EPS)
+        data *= self.new_range
+        data += self.new_min
```
olliethomas, line 178: I know this is an inplace operation, but this doesn't feel good

olliethomas: I'd much prefer if `data` was returned

MylesBartlett: How would this work with the `col_indexes`, the specification of which would lead to a subset of the data being returned?

olliethomas: will `data[:, self.col_indexes] = self._transform(data[:, self.col_indexes])` work?

MylesBartlett: I suppose it would though I'd rather keep it as it is to enforce inplaceness and thus avoid unnecessary copies of the data after the clone.

olliethomas: in that case, could the operations be made to be the explicit inplace ones e.g.
```
data.sub_(self.orig_min)
data.div_(self.orig_range.clamp_min(self._EPS))
etc
```

thomkeh (approved):

PR #51, opened 2021-08-13
Audio dataset base class implementation

DavidHurst:


MylesBartlett (approved):

bolts/data/datasets/audio/base.py
```diff
@@ -1,11 +1,67 @@
...
+        # Infer the appropriate audio-loading backend based on the operating system.
+        self.al_backend: AudioLoadingBackend = infer_al_backend()
+        self.log(f'Using {self.al_backend} as backend for audio-loading')
+        torchaudio.set_audio_backend(self.al_backend)
+
+    def __repr__(self) -> str:
+        head = "Dataset " + self.__class__.__name__
+        body = [f"Number of datapoints: {len(self)}"]
+        body.append(f"Base audio-directory location: {self.audio_dir.resolve()}")
+        body += self.extra_repr().splitlines()
```
olliethomas, line 54:
```suggestion
        body = [
            f"Number of datapoints: {len(self)}",
            f"Base audio-directory location: {self.audio_dir.resolve()}",
            *self.extra_repr().splitlines()
        ]
```

tests/data_test.py
```diff
@@ -30,9 +30,13 @@ def test_datasets(ds_cls: type[VisionDataset]) -> None:
...
 def test_audio_dataset() -> None:
-    """Tests basic functionality of an audio dataset."""
-    dataset = PBAudioDataset()
+    """Tests basic functionality of an audio dataset base class."""
+    x = torch.rand(1, 10)
+    audio_dir = Path(r"Sample path")
+    dataset = PBAudioDataset(x=x, audio_dir=audio_dir)
+
     assert dataset is not None
+    assert len(dataset) == len(x)
```
olliethomas, line 40: Maybe something like
```suggestion
    assert str(dataset).splitlines()[0] == "Dataset PBAudioDataset"
    assert str(dataset).splitlines()[1].strip() == "Number of datapoints: 1"
```
to test the repr method

bolts/data/datasets/utils.py
```diff
@@ -108,6 +112,23 @@ def img_to_tensor(img: Image.Image | np.ndarray) -> Tensor:
...
+def infer_al_backend() -> AudioLoadingBackend:
+    """Infer which audio-loading backend to use based on the operating system."""
+    return 'soundfile' if platform.system() == 'Windows' else 'sox_io'
+
+
+def apply_waveform_transform(waveform: Tensor, *, transform: AudioTform | None) -> Tensor:
+    waveform_ = waveform
+    if transform is not None:
+        waveform_ = transform(waveform_)
+    return waveform_
```
olliethomas, line 129: although there's not a hard rule about it, usually the trailing `_` is for when you want to use a python reserved word, like `class_`. But in this case, I think it's not needed.
```suggestion
   return waveform if transform is None else transform(waveform)
```

bolts/data/datasets/utils.py
```diff
@@ -108,6 +112,23 @@ def img_to_tensor(img: Image.Image | np.ndarray) -> Tensor:
...
 
 
+AudioLoadingBackend = Literal["sox_io", "soundfile"]
+
+AudioTform = Union[TF_audio.Spectrogram, TF_audio.MelSpectrogram]
+
+
+def infer_al_backend() -> AudioLoadingBackend:
+    """Infer which audio-loading backend to use based on the operating system."""
+    return 'soundfile' if platform.system() == 'Windows' else 'sox_io'
```
olliethomas, line 122: Nice 😎

pyproject.toml
```diff
@@ -28,8 +28,11 @@ lightning-bolts = "^0.3.4"
 palkit = "^0.3.8"
 torch = "*"
 torchvision = "*"
+torchaudio = "*"
 gdown = "^3.13.0"
 opencv-python = "^4.5.3"
+soundfile = {version = "*", markers = "platform_system == 'Windows'"}
+sox = {version = "*", markers = "platform_system == 'Linux' or platform_system == 'macOS'"}
```
olliethomas, line 35: I've not had to use this before with poetry, well done for working it out. 
```suggestion
soundfile = {version = "*", markers = "platform_system == 'Windows'"}
sox = {version = "*", markers = "platform_system == 'Linux' or platform_system == 'Darwin'"}
```

olliethomas (changes requested):

bolts/data/datasets/audio/base.py
```diff
@@ -1,11 +1,67 @@
...
+        lines = [head] + [" " * self._repr_indent + line for line in body]
+        return '\n'.join(lines)
+
+    def load_waveform(self, index: int) -> Tensor:
+        return torchaudio.load(self.audio_dir / self.x[index])
 
-    def __init__(self) -> None:
-        pass
+    @implements(PBDataset)
+    def _sample_x(self, index: int) -> Tensor:
```
MylesBartlett, line 64: Even though x is always a tensor, the method's signature needs to include coerce_to_tensor as this is set to true when computing the shape of x in the superclass.

bolts/data/datasets/utils.py
```diff
@@ -108,6 +112,23 @@ def img_to_tensor(img: Image.Image | np.ndarray) -> Tensor:
     )
 
 
+AudioLoadingBackend = Literal["sox_io", "soundfile"]
+
+AudioTform = Union[TF_audio.Spectrogram, TF_audio.MelSpectrogram]
```
MylesBartlett, line 117: There's no base class for torch audio transforms but they do all have the common property of taking in a tensor as input and returning a tensor. You can annotate this using typing.Callable.

bolts/data/datasets/utils.py
```diff
@@ -19,6 +20,7 @@
     np_str_obj_array_pattern,
     string_classes,
 )
+import torchaudio.transforms as TF_audio
```
MylesBartlett, line 23: It should just be T here (the F in the import below stands for 'functional'). That said, T_audio is still quite cumbersome for an alias; maybe AT and we can change the image version to IT or maybe just T as I don't think there will be any collisions.

bolts/data/datasets/utils.py
DavidHurst: Ahh ok, I know there's a lot of unspoken rules in python, good to learn some more.

tests/data_test.py
DavidHurst: Good addition, I hadn't thought to test the repr.

pyproject.toml
DavidHurst: Ahh I couldn't find what macOS would've been called system wise, shoulda asked but yea thanks wasn't took tricky.

bolts/data/datasets/utils.py
DavidHurst: I've gone with just T since the image import is for the functional module as you said.

bolts/data/datasets/utils.py
DavidHurst: Ah yes, that is much clearer I didn't know about the Callable, fixed now.

bolts/data/datasets/audio/base.py
DavidHurst: Ahh ok, yea I had just assumed since it's always a tensor there's no need.  Fixed, will check on things like this in future.

MylesBartlett (approved):

PR #52, opened 2021-08-13
fix some minor things

olliethomas:


bolts/data/datasets/vision/celeba.py
```diff
@@ -149,11 +148,11 @@ def __init__(
         )
 
         x = np.array(attrs.index)
-        s = torch.as_tensor(attrs[subclass.name].to_numpy())
-        y = torch.as_tensor(attrs[superclass.name].to_numpy())
+        _s = torch.as_tensor(attrs[subclass.name].to_numpy())
```
MylesBartlett, line 151: Why?

bolts/data/datasets/vision/celeba.py
```diff
@@ -149,11 +148,11 @@ def __init__(
...
+        _y = torch.as_tensor(attrs[superclass.name].to_numpy())
```
MylesBartlett, line 152: Why?

bolts/data/datasets/vision/isic.py
```diff
@@ -229,7 +229,7 @@ def _remove_uncertain_diagnoses(metadata_df: pd.DataFrame) -> pd.DataFrame:
             metadata_df["meta.clinical.benign_malignant"].isin({"benign", "malignant"})
         ]  # throw out unknowns
         malignant_mask = labels_df["meta.clinical.benign_malignant"] == "malignant"
-        labels_df["malignant"] = malignant_mask.astype(np.uint8)
+        labels_df["malignant"] = malignant_mask.astype("int8")
```
MylesBartlett, line 232: Do you have a good reason for this? I like to avoid string representations of things wherever possible - they're not safe.

bolts/data/datasets/vision/nico.py
```diff
@@ -111,7 +111,8 @@ def _extract_metadata(self) -> None:
         metadata = self._label_encode_metadata(metadata)
         metadata.to_csv(self._metadata_path)
 
-    def _label_encode_metadata(self, metadata: pd.DataFrame) -> pd.DataFrame:
+    @staticmethod
+    def _label_encode_metadata(metadata: pd.DataFrame) -> pd.DataFrame:
```
MylesBartlett, line 115: Fixed this in my tabular ds PR

bolts/fair/models/fair_mixup.py
```diff
@@ -24,6 +24,11 @@
 __all__ = ["FairMixup"]
 
 
+class Criterion(Protocol):
+    def __call__(self, input: Tensor, target: Tensor) -> Tensor:
```
MylesBartlett, line 29: Not sure this was necessary as there's nothing complex going on in this callable.

.github/workflows/ci.yml
```diff
@@ -38,10 +38,10 @@ jobs:
     - name: Format with isort
       run: |
         isort --check --diff .
-    - name: Setup node.js (for pyright)
-      uses: actions/setup-node@v1
-      with:
-        node-version: "12"
+#    - name: Setup node.js (for pyright)
```
MylesBartlett, line 41: Can we use pyright like we do now in palkit?

bolts/data/datasets/vision/celeba.py
olliethomas: because then we would have easily spotted that `y` wasn't used

bolts/data/datasets/vision/celeba.py
```diff
@@ -67,11 +67,11 @@ class CelebASplit(Enum):
 class CelebA(PBVisionDataset):
     """CelebA dataset."""
 
-    """The data is downloaded to `download_dir` / `_BASE_FOLDER`."""
+    # The data is downloaded to `download_dir` / `_BASE_FOLDER`.
```
MylesBartlett, line 70: Why do you prefer strings to comments?

bolts/data/datasets/vision/isic.py
olliethomas: I don't have a great reason. I agree that `np.uint8` is better, but [in the pandas docs they provide the numpy type as a string](https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.astype.html), so I assume that this is the "correct" way to do it. But 🤷‍♂️ .

The type should at least be `uint8` though, not `int8` as I've put.

bolts/fair/models/fair_mixup.py
olliethomas: it's only there for the kwargs

.github/workflows/ci.yml
olliethomas: only if you are willing to make it pass in the first place.

bolts/data/datasets/vision/celeba.py
MylesBartlett: Okay, fair enough, but maybe more meaningful pairs of names? Something like s_rad/y_rad and s_bern/y_bern.

bolts/data/datasets/vision/celeba.py
olliethomas: I have no preference, but PEP8 does. `"""` is for docstrings, and `#` is for comments

bolts/data/datasets/vision/celeba.py
olliethomas: I also don't like reassigning variables if it can be helped

bolts/data/datasets/vision/isic.py
MylesBartlett: Let's keep it as it as np.uint8 then it your only reason is the pandas docs - we follow our own philosophy.

bolts/data/datasets/vision/nico.py
MylesBartlett: But we can fix it here as well

bolts/fair/models/fair_mixup.py
MylesBartlett: But the signature doesn't include kwargs?

.github/workflows/ci.yml
MylesBartlett: Oh right, there's that problem with the sample classes

bolts/data/datasets/vision/celeba.py
MylesBartlett: Aren't these comments and not docstrings?

bolts/fair/models/fair_mixup.py
olliethomas: as just a callable, this line was giving a type error: `criterion(input=pred, target=tgt_a)` as it didn't expect kwargs. Now it knows that input and target are the names.

MylesBartlett: Oh I see. Sorry I thought you meant **kwargs which is one reason to use Protocol over callable.

bolts/data/datasets/vision/celeba.py
olliethomas: https://www.python.org/dev/peps/pep-0008/#comments

MylesBartlett: Okay?

MylesBartlett: Oh sorry I read the change the wrong way round...

MylesBartlett (approved):

PR #53, opened 2021-08-19
Ecoacoustics dataset

DavidHurst:


tests/data_test.py
```diff
@@ -30,15 +32,48 @@ def test_datasets(ds_cls: type[VisionDataset]) -> None:
...
     x = torch.rand(1, 10)
     audio_dir = Path(r"Sample path")
     dataset = PBAudioDataset(x=x, audio_dir=audio_dir)
 
     assert dataset is not None
     assert len(dataset) == len(x)
+
+    
+def test_audio_dataset() -> None:
+    root_dir = Path("~/Data").expanduser()
```
thomkeh, line 45: this only works on your machine, right?

bolts/data/datasets/audio/ecoacoustics_ds.py
```diff
@@ -0,0 +1,121 @@
...
+    """Dataset for audio data collected in various geographic locations."""
+
+    LABELS_DIR: ClassVar[str] = "AvianID_AcousticIndices"
+    METADATA_FILENAME: ClassVar[str] = "metadata.csv"
+
+    _FILE_INFO: ClassVar[FileInfo] = FileInfo(name="EcoacousticsDS.zip", id="PLACEHOLDER")
+    _BASE_FOLDER: ClassVar[str] = "EcoacousticsDS"
+    _EC_LABELS_FILENAME: ClassVar[str] = "EC_AI.csv"
+    _UK_LABELS_FILENAME: ClassVar[str] = "UK_AI.csv"
+
```
MylesBartlett, line 38:
```suggestion
@parsable
```
importing `parsable` from `kit`

bolts/data/datasets/audio/ecoacoustics_ds.py
```diff
@@ -0,0 +1,121 @@
...
+    def _label_encode_metadata(self, metadata: pd.DataFrame) -> pd.DataFrame:
+        """Label encode the extracted concept/context/superclass information."""
+        for col in metadata.columns:
+            # Skip over filepath and filename columns
+            if "file" not in col:
+                # Add a new column containing the label-encoded data
+                metadata[f"{col}_le"] = metadata[col].factorize()[0]
+        return metadata
+
+    def _extract_metadata(self, target_attr) -> None:
```
MylesBartlett, line 91: Since this is only run if a metadata file found to not already exist, you should just extract the metadata for all attributes; as it is at the moment, as I understand it, changing the attribute will result in an error (though whether the target attribute actually needs to be specified is another point of discussion...).

bolts/data/datasets/audio/ecoacoustics_ds.py
```diff
@@ -0,0 +1,121 @@
...
+from typing_extensions import Literal
+
+from bolts.data.datasets.audio.base import PBAudioDataset
+from bolts.data.datasets.utils import AudioTform, FileInfo
+
+__all__ = ["EcoacousticsDS"]
+SoundscapeAttr = Literal["habitat", "site"]
+
+
+class EcoacousticsDS(PBAudioDataset):
```
MylesBartlett, line 28:
```suggestion
class Ecoacoustics(PBAudioDataset):
```
I think the `DS` suffix can be dropped -- the fact the class lives in `.datasets` makes it clear enough that it is a dataset.

tests/data_test.py
olliethomas: I think this is copied from the other tests which are marked as slow so that they are skipped by the CI, but you can run locally

bolts/data/datasets/audio/__init__.py
```diff
@@ -1 +1 @@
-from .ecoacoustics import *
+from .ecoacoustics_ds import *
```
olliethomas, line 1: the filename can drop `_ds` too

bolts/data/datasets/audio/ecoacoustics_ds.py
```diff
@@ -0,0 +1,121 @@
...
+        ec_labels = pd.read_csv(self.ec_labels_path, encoding="ISO-8859-1")
+        uk_labels = pd.read_csv(self.uk_labels_path, encoding="ISO-8859-1")
+        all_labels = pd.concat([ec_labels, uk_labels])
+
+        target_attrs: list[SoundscapeAttr] = []
+
+        for f_name in filenames:
+            matched_row = all_labels.loc[all_labels['fileName'] == f_name]
+            target_attrs.append(np.nan if matched_row.empty else matched_row.iloc[0][target_attr])
+        target_attrs = pd.Series(target_attrs)
```
MylesBartlett, line 111:
```suggestion
        metadata = metadata.merge(all_labels, how="left")
```
Could the same be achieved with `pd.merge` (as above)?

bolts/data/datasets/audio/ecoacoustics_ds.py
```diff
@@ -0,0 +1,121 @@
...
+        s = None
+
+        super().__init__(x=x, y=y, s=s, transform=transform, audio_dir=self._base_dir)
+
+    def _files_unzipped(self) -> bool:
+        dir = self.root / self._BASE_FOLDER
+        zip_checks = [
+            zipfile.is_zipfile(join(dir, f)) for f in listdir(dir) if isfile(join(dir, f))
+        ]
+        return False if True in zip_checks else True
```
olliethomas, line 80:
```suggestion
        return True not in zip_checks
```

tests/data_test.py
```diff
@@ -30,15 +32,48 @@ def test_datasets(ds_cls: type[VisionDataset]) -> None:
...
+
+    ds_cls_dnwld = EcoacousticsDS(root=root_dir, target_attr=target_attribute)
+    assert ds_cls_dnwld is not None
+
+    ds_cls_no_dnwld = EcoacousticsDS(root=root_dir, download=False, target_attr=target_attribute)
+    assert ds_cls_no_dnwld is not None
+
+    # Test __len__
+    num_audio_samples = []
+    num_audio_samples.extend(root_dir.glob("**/*.wav"))
```
olliethomas, line 57:
```suggestion
    num_audio_samples = list(root_dir.glob("**/*.wav"))
```

tests/data_test.py
```diff
@@ -30,15 +32,48 @@ def test_datasets(ds_cls: type[VisionDataset]) -> None:
...
+    """Tests basic functionality of the base audio dataset class."""
     x = torch.rand(1, 10)
     audio_dir = Path(r"Sample path")
     dataset = PBAudioDataset(x=x, audio_dir=audio_dir)
 
     assert dataset is not None
     assert len(dataset) == len(x)
+
+    
+def test_audio_dataset() -> None:
```
olliethomas, line 44: can't be named the same as another test.

olliethomas: Add `@pytest.mark.slow`

bolts/data/datasets/audio/ecoacoustics_ds.py
```diff
@@ -0,0 +1,121 @@
...
+        self.metadata = pd.read_csv(self._base_dir / "metadata.csv")
+
+        x = self.metadata["filePath"].to_numpy()
+        y = torch.as_tensor(self.metadata[f'{target_attr}_le'])
+        s = None
+
+        super().__init__(x=x, y=y, s=s, transform=transform, audio_dir=self._base_dir)
+
+    def _files_unzipped(self) -> bool:
+        dir = self.root / self._BASE_FOLDER
```
olliethomas, line 76:
```suggestion
        dir_ = self.root / self._BASE_FOLDER
```
`dir` mirrors a python builtin

bolts/data/datasets/audio/ecoacoustics_ds.py
```diff
@@ -0,0 +1,121 @@
...
+        download: bool = True,
+        target_attr: SoundscapeAttr = "habitat",
+        transform: Optional[AudioTform] = None,
+    ) -> None:
+
+        self.root = Path(root)
+        self.download = download
+        self._base_dir = self.root / self._BASE_FOLDER
+        self.labels_dir = self._base_dir / self.LABELS_DIR
+        self._metadata_path = self._base_dir / "metadata.csv"
```
olliethomas, line 52: maybe create a constant rather than keep using the `"metadata.csv"` string directly

bolts/data/datasets/audio/ecoacoustics_ds.py
```diff
@@ -0,0 +1,121 @@
...
+        *,
+        download: bool = True,
+        target_attr: SoundscapeAttr = "habitat",
+        transform: Optional[AudioTform] = None,
+    ) -> None:
+
+        self.root = Path(root)
+        self.download = download
+        self._base_dir = self.root / self._BASE_FOLDER
+        self.labels_dir = self._base_dir / self.LABELS_DIR
```
olliethomas, line 51: to avoid confusion with the `LABELS_DIR` constant, one of these should be renamed

bolts/data/datasets/audio/ecoacoustics_ds.py
```diff
@@ -0,0 +1,121 @@
+"""Ecoacostics dataset provided A. Eldridge et al.
+    Alice Eldridge, Paola Moscoso, Patrice Guyot, & Mika Peck. (2018).
+    Data for "Sounding out Ecoacoustic Metrics: Avian species richness
+    is predicted by acoustic indices in temperate but not tropical
+    habitats" (Final) [Data set].
+    Zenodo. https://doi.org/10.5281/zenodo.1255218
+"""
+from __future__ import annotations
+import os
```
olliethomas, line 9: instead of importing `os`, i think you only use `os.sep`, so the line below can also import this

olliethomas (changes requested):

bolts/data/datasets/audio/__init__.py
MylesBartlett: Good catch!

bolts/data/datasets/audio/ecoacoustics_ds.py
MylesBartlett: I think we can avoid usnig `os.sep` in this case by using pandas string methods.

tests/data_test.py
DavidHurst: Ah I didn't think it fully through, I thought It just opens the Data dir on the machine, yes copied from other tests, will this be a problem?

bolts/data/datasets/audio/ecoacoustics_ds.py
DavidHurst: Yes this does need work but I thought improvement should be guided on how we want to handle multi-class.

bolts/data/datasets/utils.py
```diff
@@ -20,6 +20,7 @@
     np_str_obj_array_pattern,
     string_classes,
 )
+import torchaudio.transforms as TF_audio
```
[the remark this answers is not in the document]
DavidHurst: Thought I already changed this smh, good spot.

bolts/data/datasets/audio/ecoacoustics_ds.py
DavidHurst: Good point, bit redundant, I just couldn't think of a catchy name like "NICO" or "MNIST" lol, will fix

DavidHurst: Fixed

bolts/data/datasets/audio/ecoacoustics_ds.py
DavidHurst: Looks like it can, solves problem of erroring out if a different target attribute is selected as well.

bolts/data/datasets/audio/__init__.py
DavidHurst: ^^ All changed and tested, works renamed.

bolts/data/datasets/audio/ecoacoustics_ds.py
DavidHurst: Pretty much solved by merging the all labels file and metadata file.

bolts/data/datasets/audio/ecoacoustics_ds.py
DavidHurst: Facepalm lol, not sure this function is even really valid anyway tho? Should probably check for the main files only or is checking that everything is unzipped ok?

tests/data_test.py
DavidHurst: Much nicer, I'm relying too much on the copy paste but was also lazy not checking the return of `glob` thought it might have necessitated extend by being a list or something.

tests/data_test.py
DavidHurst: I didn't even realise I did this wow lol, I might just remove the test for the base class? Doesn't do much?

bolts/data/datasets/audio/ecoacoustics_ds.py
DavidHurst: The more you know. Thanks.

DavidHurst: A better IDE would've probably told me this as well...

bolts/data/datasets/audio/ecoacoustics_ds.py
DavidHurst: I actually already had a constant I just wasn't using it wow lol, good spot.

bolts/data/datasets/audio/ecoacoustics_ds.py
DavidHurst: Good idea.

tests/data_test.py
olliethomas: not a problem, just mark as slow

bolts/data/datasets/audio/ecoacoustics.py
```diff
@@ -1 +1,113 @@
...
+        return metadata
+
+    def _extract_metadata(self, target_attr) -> None:
+        self.log("Extracting metadata.")
+        waveform_paths = list(self._base_dir.glob("**/*.wav"))
+
+        # Extract filepaths and names.
+        waveform_paths_str = [str(wvfrm.relative_to(self._base_dir)) for wvfrm in waveform_paths]
+        filepaths = pd.Series(waveform_paths_str)
+        pd.Series([path_str.split(os.sep)[-1] for path_str in waveform_paths_str])
```
olliethomas, line 98: os not imported so os.sep won't work. If you used waveform_paths instead of waveform_path_str, then you could just do `pd.Series([pth.name for pth in waveform_paths])` using [Path.name](https://docs.python.org/3/library/pathlib.html#pathlib.PurePath.name)

DavidHurst: Myles recommended I replace all that splitting stuff wit pandas string functions which I have done, no os.sep or os at all. might be in ecoacoustics-preprocessing commits but is definitely fixed.

PR #61, opened 2021-09-06
Mostly Linter Things

olliethomas:
- mostly linter things
- balck and isort


conduit/models/self_supervised/multicrop.py
```diff
@@ -51,7 +49,9 @@ def with_dino_transform(
         local_crops_scale: tuple[float, float] = (0.05, 0.4),
         local_crops_number: int = 8,
     ) -> MultiCropTransform:
-        from conduit.models.self_supervised.dino.transforms import dino_train_transform
+        from conduit.models.self_supervised.dino.transforms import (  # TODO: Cyclic Import
```
MylesBartlett, line 52: This doesn't matter because the import is lazy

conduit/models/self_supervised/multicrop.py
```diff
@@ -51,7 +49,9 @@ def with_dino_transform(
...
+            dino_train_transform,
+        )
```
olliethomas, line 54:
```suggestion
        from conduit.models.self_supervised.dino.transforms import dino_train_transform
```

conduit/models/self_supervised/moco/transforms.py
```diff
@@ -11,31 +9,12 @@
...
 
 __all__ = [
-    "GaussianBlur",
     "moco_ft_transform",
     "moco_test_transform",
     "mocov2_train_transform",
 ]
 
-
-class GaussianBlur:
```
MylesBartlett, line 21: I don't agree with this move. While this transform is used by DINO that so happens because it inherits its transforms partly from MoCo.

conduit/models/self_supervised/dino/transforms.py
```diff
@@ -9,11 +9,12 @@
 from conduit.constants import IMAGENET_STATS
 from conduit.data.datasets.utils import PillowTform
 from conduit.data.structures import MeanStd
-from conduit.models.self_supervised.moco import GaussianBlur
 from conduit.models.self_supervised.multicrop import MultiCropTransform
 
 __all__ = ["Solarization", "dino_train_transform"]
 
+from conduit.models.self_supervised.transforms import GaussianBlur
```
MylesBartlett, line 16: Misplaced

conduit/models/self_supervised/dino/loss.py
```diff
@@ -107,11 +106,11 @@ def forward(
         total_loss = student_logits_seq[-1].new_zeros(())
         n_loss_terms = student_logits_seq[-1].new_zeros(())
         for iq, q in enumerate(teacher_probs_seq):
-            for v in range(len(student_logits_seq)):
+            for v, sl in enumerate(student_logits_seq):
```
MylesBartlett, line 109: bit of a pointless change

conduit/fair/metrics/accuracy.py
```diff
@@ -10,7 +10,7 @@ class FbAccuracy(Accuracy):
 
     @property
     def __name__(self) -> str:
```
MylesBartlett, line 12: This metric is no longer used by the way

conduit/data/datasets/vision/celeba.py
```diff
@@ -10,7 +10,8 @@
 import torch
 
 from conduit.data.datasets.utils import FileInfo, ImageTform, download_from_gdrive
-from conduit.data.datasets.vision.base import CdtVisionDataset
+
+from .base import CdtVisionDataset
```
MylesBartlett, line 14: Why make this change? The absolute path is perfectly functional and is what auto-import uses.

conduit/architectures/vit.py
```diff
@@ -202,6 +201,8 @@ def __init__(
         norm_layer: type[nn.LayerNorm] | type[nn.BatchNorm1d] = nn.LayerNorm,
     ) -> None:
         super().__init__()
+        if isinstance(img_size, int):
```
thomkeh, line 204: Hm, maybe check for "not list or tuple" instead?

conduit/callbacks/posthoc_eval.py
```diff
@@ -22,6 +22,8 @@ class PostHocEval(pl.Callback):
         - ``self.trainer''
     """
 
+    eval_clf: FineTuner
```
MylesBartlett, line 25: let's just delete this file

conduit/data/datasets/vision/celeba.py
olliethomas: it's clearer that you're importing from a file that's in the same module

thomkeh: Local imports are less likely to be cyclic

conduit/models/self_supervised/moco/transforms.py
olliethomas: but there's nothing moco specific about this transform, and importing from another module at the same level isn't great

thomkeh (approved):

conduit/data/datasets/vision/celeba.py
MylesBartlett: But such imports are also more brittle in the face of refactoring

conduit/models/self_supervised/dino/transforms.py
```diff
@@ -10,11 +10,10 @@
 from conduit.data.datasets.utils import PillowTform
 from conduit.data.structures import MeanStd
 from conduit.models.self_supervised.multicrop import MultiCropTransform
+from conduit.models.self_supervised.transforms import GaussianBlur
 
 __all__ = ["Solarization", "dino_train_transform"]
 
-from conduit.models.self_supervised.transforms import GaussianBlur
```
MylesBartlett: Still think this should stay in MoCo

conduit/models/self_supervised/dino/transforms.py
```diff
@@ -10,11 +10,10 @@
 from conduit.data.datasets.utils import PillowTform
 from conduit.data.structures import MeanStd
 from conduit.models.self_supervised.multicrop import MultiCropTransform
+from conduit.models.self_supervised.transforms import GaussianBlur
```
MylesBartlett, line 13: Still think this should stay in moco

conduit/models/self_supervised/moco/transforms.py
MylesBartlett: Except for the fact that the implementation and default parameterisation matches that in the official MoCo implementation - different libraries implement the Gaussian kernel in substantially different ways (in the way that the radius is computed, for instance)

MylesBartlett (approved):

PR #63, opened 2021-09-07
Pad spectrograms terminally with zeros if not integer-divisible by specgram_segment_len

MylesBartlett:


conduit/data/datasets/audio/ecoacoustics.py
```diff
@@ -275,15 +275,17 @@ def _preprocess_audio(self) -> None:
             waveform, sr = torchaudio.load(path)
             waveform = F.resample(waveform, sr, self.resample_rate)
             specgram = to_specgram(waveform)
+            audio_len = waveform.size(-1) / self.resample_rate
+            remainder = audio_len % self.specgram_segment_len
+            if remainder:
+                padding = torch.zeros(
```
olliethomas, line 281: is there a way to do this with zeros_like?

olliethomas:
if the remainder is 1, won't this produce a really weird sample?

conduit/data/datasets/audio/ecoacoustics.py
MylesBartlett: No because we don't want the last dimension to match in size; there's no cuda-involvement here so the like/new factory methods aren't necessary.

MylesBartlett:
What do you suggest?

olliethomas:
maybe if under 50% of a segment size, then discard, else pad?

olliethomas (approved):

PR #68, opened 2021-09-08
Alternative fix for the segmenting problem

thomkeh:


conduit/data/datasets/audio/ecoacoustics.py
```diff
@@ -302,7 +304,10 @@ def _preprocess_audio(self) -> None:
                     :, : num_segments * self.specgram_segment_len * self.resample_rate
                 ]
 
-            waveform_segments = waveform.chunk(chunks=num_segments, dim=-1)
-            for i, segment in enumerate(waveform_segments):
-                specgram = to_specgram(segment)
-                torch.save(specgram, f=self._processed_audio_dir / f"{waveform_filename}={i}.pt")
+            # because of `center=True`, the spectrogram adds an additional frame
```
olliethomas, line 307: this seems brittle

thomkeh: Using `torch.chunk` seem brittle in general. The correct way is to compute how long each segment should be and then cut the spectrogram accordingly.

thomkeh:
@olliethomas , see latest commit. I replaced chunk with split.

thomkeh:
on reflection, I don't really care about this

PR #69, opened 2021-09-08
make explicit that the transform should be passed as an arg

olliethomas:


conduit/data/datasets/audio/base.py
```diff
@@ -65,4 +65,4 @@ def load_waveform(self, index: int) -> Tensor:
     @implements(CdtDataset)
     def _sample_x(self, index: int, *, coerce_to_tensor: bool = False) -> Tensor:
         waveform = self.load_waveform(index)
-        return apply_waveform_transform(waveform)
+        return apply_waveform_transform(waveform, transform=None)
```
olliethomas, line 68: i don't think that this is used yet, but it was missing a required kwarg

conduit/data/datasets/audio/ecoacoustics.py
```diff
@@ -98,8 +98,6 @@ def __init__(
         transform: Optional[AudioTform] = None,
         resample_rate: int = 22050,
         specgram_segment_len: float = 15,
-        num_freq_bins: int = 120,
-        hop_length: int = 60,
```
olliethomas, line 102: these are transform args, so should be provided as part of the transform

conduit/data/datasets/audio/ecoacoustics.py
```diff
@@ -117,9 +115,9 @@ def __init__(
         self.specgram_segment_len = specgram_segment_len
         self.resample_rate = resample_rate
         self._n_sgram_segments = int(self._AUDIO_LEN / specgram_segment_len)
-        self.hop_length = hop_length
-        self.num_freq_bins = num_freq_bins
-        self.preprocess_transform = T.Spectrogram(n_fft=num_freq_bins, hop_length=hop_length)
+        if transform is None:
+            transform = T.Spectrogram(n_fft=120, hop_length=60)
+        self.transform = transform
```
olliethomas, line 120: needs setting here as used in preprocess

conduit/data/datasets/audio/ecoacoustics.py
```diff
@@ -299,10 +295,11 @@ def _preprocess_audio(self) -> None:
                     f"(fractional remainder must be greater than 0.5): discarding terminal segment."
                 )
                 waveform = waveform[
-                    :, : num_segments * self.specgram_segment_len * self.resample_rate
+                    :, : int(num_segments * self.specgram_segment_len * self.resample_rate)
```
olliethomas, line 298: specgram_segment_len is of type float

conduit/data/datasets/utils.py
```diff
@@ -122,7 +122,9 @@ def img_to_tensor(img: Image.Image | np.ndarray) -> Tensor:
 
 def infer_al_backend() -> AudioLoadingBackend:
     """Infer which audio-loading backend to use based on the operating system."""
-    return 'soundfile' if platform.system() == 'Windows' else 'sox_io'
+    soundfile: Literal["soundfile"] = "soundfile"
+    sox: Literal["sox_io"] = "sox_io"
+    return soundfile if platform.system() == "Windows" else sox
```
olliethomas, line 127: my ide can't deal with returning a str without being of literal type

MylesBartlett: That's a bit shit

conduit/data/datasets/audio/base.py
MylesBartlett: there is a transform attribute, isn't there?

olliethomas: because the preprocessing is no longer done on the fly, I think that this should now be
```suggestion
        return self.load_spectrogram(path)
```

olliethomas: or index, whatever

MylesBartlett: There is no such method as `load_spectrogram` currently, btw, right?

conduit/data/datasets/audio/ecoacoustics.py
```diff
@@ -299,10 +295,11 @@ def _preprocess_audio(self) -> None:
...
                 ]
 
             waveform_segments = waveform.chunk(chunks=num_segments, dim=-1)
+            assert self.transform is not None
```
MylesBartlett, line 302: transform shouldn't be optional, or at least, if it is, the preprocessing shouldn't be run and the original wav files should be used.

MylesBartlett (approved):

conduit/data/datasets/audio/ecoacoustics.py
thomkeh: maybe make `transform` a required argument? David seemed to think you want to hard-code the transform now.

conduit/data/datasets/utils.py
```diff
@@ -125,7 +125,9 @@ def img_to_tensor(img: Image.Image | np.ndarray) -> Tensor:
 
 def infer_al_backend() -> AudioLoadingBackend:
     """Infer which audio-loading backend to use based on the operating system."""
-    return 'soundfile' if platform.system() == 'Windows' else 'sox_io'
+    soundfile: Literal["soundfile"] = "soundfile"
+    sox: Literal["sox_io"] = "sox_io"
```
thomkeh, line 129: This should also work to make the IDE see them as `Literal` (at least this works in mypy)

```suggestion
    soundfile: Final = "soundfile"
    sox: Final = "sox_io"
```

conduit/data/datasets/audio/ecoacoustics.py
thomkeh: maybe rename it to `preprocessing` as well

conduit/data/datasets/audio/ecoacoustics.py
```diff
@@ -113,9 +111,11 @@ def __init__(
         self.specgram_segment_len = specgram_segment_len
         self.resample_rate = resample_rate
         self._n_sgram_segments = int(self._AUDIO_LEN / specgram_segment_len)
-        self.hop_length = hop_length
-        self.num_freq_bins = num_freq_bins
-        self.preprocess_transform = T.Spectrogram(n_fft=num_freq_bins, hop_length=hop_length)
+        if preprocessing_transform is None:
+            preprocessing_transform = T.Spectrogram(n_fft=120, hop_length=60)
+        self.transform = (
```
MylesBartlett, line 116:
```suggestion
        self.transform = (
```
Rename this to `preprocessing_transform`

PR #70, opened 2021-09-08
Add subset-factory function for CdtDataset instances.

MylesBartlett:


conduit/data/datasets/utils.py
```diff
@@ -270,8 +270,8 @@ def _subset_from_indices_(_indices: list[int] | slice) -> None:
         if subset.s is not None:
             subset.s = subset.s[_indices]
 
-    if old_indices is not None:
-        _subset_from_indices_(old_indices)
+    if current_indices is not None:
+        _subset_from_indices_(current_indices)
     if indices is not None:
         _subset_from_indices_(indices)
```
olliethomas: I don't really like this pattern. Could `_subset_from_indices` take subset as an arg and return `subset`?

olliethomas (approved):

PR #71, opened 2021-09-08
Implemented datamodule for ecoacoustics dataset.

DavidHurst:


olliethomas:
there should be a test

conduit/data/datasets/audio/base.py
```diff
@@ -4,7 +4,7 @@
 from kit import implements
 import numpy as np
 import numpy.typing as npt
-from torch import Tensor
+from torch import Tensor, load
```
MylesBartlett, line 7:
```suggestion
from torch import Tensor
```

conduit/data/datasets/audio/base.py
```diff
@@ -59,10 +59,11 @@ def __repr__(self) -> str:
         lines = [head] + [" " * self._repr_indent + line for line in body]
         return '\n'.join(lines)
 
-    def load_waveform(self, index: int) -> Tensor:
-        return torchaudio.load(self.audio_dir / self.x[index])
+    def load_sample(self, index: int) -> Tensor:
+        path = self.audio_dir / self.x[index]
+        return torchaudio.load(path) if str(self.x[index]).endswith('.wav') else load(path)
```
MylesBartlett, line 64:
```suggestion
        return torchaudio.load(path) if str(self.x[index]).endswith('.wav') else torch.load(path)
```

conduit/data/datamodules/audio/ecoacoustics.py
```diff
@@ -0,0 +1,99 @@
...
+            training_mode=training_mode,
+        )
+        self.image_size = image_size
+        self.class_train_props = class_train_props
+        self.specgram_segment_len = specgram_segment_len
+        self.num_freq_bins = num_freq_bins
+
+    @property  # type: ignore[misc]
+    @implements(CdtAudioDataModule)
+    def _base_augmentations(self) -> A.Compose:
```
MylesBartlett, line 65: these augmentation properties were removed a couple of weeks ago and replaced with default_transforms, with the transforms now able to be freely specified through the init.

conduit/data/datamodules/audio/base.py
```diff
@@ -0,0 +1,89 @@
...
+
+    @implements(CdtDataModule)
+    def setup(self, stage: Stage | None = None) -> None:
+        train, val, test = self._get_splits()
+        train = ImageTransformer(train, transform=self._augmentations(train=True))
+        if self.instance_weighting:
+            train = InstanceWeightedDataset(train)
+        self._train_data = train
+        self._val_data = ImageTransformer(val, transform=self._augmentations(train=False))
+        self._test_data = ImageTransformer(test, transform=self._augmentations(train=False))
```
MylesBartlett, line 81: You'll need to adapt the ImageTransformer class into an AudioTransformer class that calls apply_waveform_transform instead of apply_image_transform.

conduit/data/datamodules/audio/base.py
```diff
@@ -0,0 +1,89 @@
...
+
+    @property
+    @abstractmethod
+    def _base_augmentations(self) -> A.Compose:
+        ...
+
+    @property
+    @abstractmethod
+    def _train_augmentations(self) -> A.Compose:
+        ...
```
MylesBartlett, line 71: As mentioned below, these properties have been removed as of a couple of weeks ago. That aside, the properties should return nn.Sequential objects, not A.Compose (which is image-specific).

conduit/data/datamodules/audio/base.py
DavidHurst: Ok will do

conduit/data/datamodules/audio/base.py
DavidHurst: Done

olliethomas:
Is this ready for review?

DavidHurst:
There is the ImageTransformer to AudioTransformer conversion to do but I've kind of put it on the backburner for now, I also need to parameterise the spectrogram transform so maybe I can do that in another branch and this can get merged?

PR #72, opened 2021-09-09
Add Waterbirds dataset.

MylesBartlett:


conduit/data/datasets/vision/waterbirds.py
```diff
@@ -0,0 +1,74 @@
...
+        self.download = download
+        if self.download:
+            download_from_url(
+                file_info=self._FILE_INFO,
+                root=self.root,
+                logger=self.logger,
+                remove_finished=True,
+            )
+        else:
+            raise FileNotFoundError(
```
thomkeh, line 56: Can this check be somehow be put into shared code?

thomkeh (approved):

conduit/data/datasets/vision/waterbirds.py
MylesBartlett: yes, I was thinking the same thing; this else block is repeated an awful lot. Nonetheless, I'm not going to do it in this PR as it's a fairly time-consuming change to make.

olliethomas (approved):

PR #81, opened 2021-09-15
Convert datamodule classes into attrs classes.

MylesBartlett:


MylesBartlett:
We need a variant of parsable that operates at the class level

thomkeh:
Does configen work on this?

MylesBartlett:
> Does configen work on this?

Yes; do you think I would have spent an hour making the switch without first testing for configen compatibility? configen doesn't work with the standard attr.s decorator but it does work with the attr.define decorator designed for pyright compatibility.

olliethomas:
> > Does configen work on this?
> 
> Yes; do you think I would have spent an hour making the switch without first testing for configen compatibility? configen doesn't work with the standard attr.s decorator but it does work with the attr.define decorator designed for pyright compatibility.

This pleasantly surprises me.

thomkeh (approved):
LGTM

PR #84, opened 2021-09-16
remove `from __future__ import annotations` so that attrs can work a bit better

olliethomas:


olliethomas:
I'm sure there will be some mistakes, but mypy is about the same as it was, and the tests pass.

MylesBartlett:
Merged with `configen` branch.

PR #87, opened 2021-09-16
Get property name from call stack for error message

thomkeh:


conduit/data/datamodules/base.py
```diff
@@ -184,7 +184,7 @@ def test_dataloader(self) -> DataLoader:
     def dims(self) -> Tuple[int, ...]:
         if self._dims:
             return self._dims
-        self._check_setup_called("size")
+        self._check_setup_called(caller="size")
```
MylesBartlett, line 187:
```suggestion
        self._check_setup_called(caller="dims")
```

MylesBartlett (approved):

MylesBartlett:
I considered something similar, determining the function name using inspect instead of hard-coding the names as strings, but, after browsing SO, thought it looked too convoluted; the way you've implemented it looks good, however.

conduit/data/datamodules/base.py
thomkeh: oh, was this not on purpose "size"?

PR #88, opened 2021-09-17
Add Relay class for orchestrating runs.

MylesBartlett:


pyproject.toml
```diff
@@ -34,6 +34,8 @@ opencv-python = "^4.5.3"
 soundfile = {version = "*", markers = "platform_system == 'Windows'"}
 sox = {version = "*", markers = "platform_system == 'Linux' or platform_system == 'macOS'"}
 attrs = "^21.2.0"
+hydra-core = {git = "https://github.com/facebookresearch/hydra.git", rev= "main"}
```
thomkeh, line 37: Why not a version?

conduit/relays/base.py
```diff
@@ -0,0 +1,184 @@
+import imp
```
thomkeh, line 1: imp seems to be deprecated?

MylesBartlett: Indeed; importlib requires 4 lines of code to do the same thing apparently, but nonetheless I have replaced imp with it.

PR #96, opened 2021-09-24
Run `poetry update` and make sure the slow tests actually pass

olliethomas:
- fix versions
- move root to fixture
- update tests a little

Just wanted to update poetry, then spotted that the tests should really have the dir for the data in a fixture.

### DIFF
--- a/tests/data_test.py
+++ b/tests/data_test.py
@@ -90,7 +90,7 @@ def test_ecouacoustics_labels(root: Path):
         preprocessing_transform=AT.Spectrogram(n_fft=120, hop_length=60),
         transform=None,
     )
-    metadata = pd.read_csv(root / ds._BASE_FOLDER / ds.METADATA_FILENAME)
+    metadata = pd.read_csv(root / ds.__class__.__name__ / ds.METADATA_FILENAME)
     # Test metadata aligns with labels file.
     audio_samples_to_check = [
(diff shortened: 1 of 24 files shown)


tests/data_test.py
```diff
@@ -1,47 +1,58 @@
...
+        assert isinstance(_ds, TernarySample)
+        assert _ds.x[0] is not None
 
 
 @pytest.mark.slow
-def test_ecoacoustics_dataset() -> None:
-    root_dir = Path("~/Data").expanduser()
+@pytest.mark.parametrize("ds_cls", [Ecoacoustics])
+def test_audio_dataset(root: Path, ds_cls: Type[Ecoacoustics]) -> None:
+    root_dir = Path(root).expanduser()
```
thomkeh, line 49:
```suggestion
    root_dir = root
```

tests/data_test.py
```diff
@@ -1,47 +1,58 @@
...
     base_dir = root_dir / "Ecoacoustics"
     target_attribute = "habitat"
```
thomkeh, line 51: If you're parameterizing this test, then this should probably be parameterized as well

olliethomas: yeah, the problem, or rather, what I assume is a problem, that I don't want to touch, is that the "preprocessing" doesn't save in specific dirs

MylesBartlett (approved):

PR #99, opened 2021-09-29
add crime dataset conf class

olliethomas:


olliethomas:
Extended this in #100

thomkeh:
All the enums seem broken

PR #103, opened 2021-09-30
Run tests once a week on `main` (as regression tests)

thomkeh:


thomkeh:
In theory this should run the tests with `--run-slow`, but I feel like that would take days to run.

MylesBartlett (approved):

PR #111, opened 2021-10-20
Rework Ecoacoustics dataset.

MylesBartlett:
The current implementation of the Ecoacoustics dataset awkwardly straddles the line between image and audio datasets  due to the audio files being converted into spectrograms and saved as .pt files during the processing step, something which also severely hinders flexibility due to being locked into a specific spectrogram parameterisation. The motivation for this was that since the audio files already needed to be segmented we may as well also do away with the overhead of an on-the-fly transformation (though how significant said overhead actually is in practice, with multiple data-loading workers enabled, was never actually determined). The current version of this PR aims to redress these two problems by simply removing the spectrogram-conversion step from preprocessing method, such that the waveforms segments are saved as .wav files and the data can thus be treated as audio and not pseudo-image data. However, with this approach there still remains inflexibility stemming from the segment length; it may be better to do no preprocessing at all and instead randomly sample segments (one from each waveform, in the name of iidness) on-the-fly using a custom transform, akin to in an image-loading pipeline, though I'm not sure how costly this would be in comparison.

### DIFF
--- a/conduit/data/datamodules/base.py
+++ b/conduit/data/datamodules/base.py
@@ -22,7 +22,7 @@ from conduit.data.datasets.utils import (
 )
 from conduit.data.datasets.wrappers import InstanceWeightedDataset
 from conduit.data.structures import ImageSize, TrainValTestSplit
-from conduit.logging import LoggingContext, init_logger
+from conduit.logging import init_logger
 from conduit.types import Stage
 
 __all__ = ["CdtDataModule"]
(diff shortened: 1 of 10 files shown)


ivorsimpson:
Thanks for looking into this and your changes. It looks like you have also added support for multiple attributes, but I haven't tested this yet. 
I'm of the opinion that fully dynamic preprocessing would be the optimal strategy. The segment length is not intrinsically a different factor of variability to the spectrogramming, and as .wav is uncompressed, I would hope it should not be too inefficient to dynamically sample segments. Moreover, it's likely that at certain parts of model training/evaluation we would take the full minute long spectrogram as input into the model, and accordingly the preprocessing would just copy the whole dataset!

PR #121, opened 2021-11-15
make the default datamodule batch sampler the pytorch default

olliethomas:


MylesBartlett:
PR is unnecessary as the desired behaviour already exists by virtue of the training_mode attribute.

PR #122, opened 2021-12-06
remove frozen aspect of dataclasses

olliethomas:
PyTorch Lightning 1.5+ exposes that frozen dataclasses shouldn't be allowed. While the data indeed does not change, they need to change the device of the Tensor, which breaks the frozen aspect.

### DIFF
--- a/conduit/data/structures.py
+++ b/conduit/data/structures.py
@@ -29,7 +29,7 @@ __all__ = [
 ]
 
 
-@dataclass(frozen=True)
+@dataclass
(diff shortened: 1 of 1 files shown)


thomkeh:
Btw, why not namedtuples? I think they're more lightweight.

thomkeh (approved):

olliethomas:
> Btw, why not namedtuples? I think they're more lightweight.

🤷‍♂️  I don't know. There's probably a reason

PR #125, opened 2022-03-10
camelyon17

MylesBartlett:
- Adapt Camelyon17 dataset from WILDS.
- Add ``__iter__`` methods to Sample classes to allow for unpacking.

### DIFF
--- a/conduit/data/datamodules/vision/camelyon17.py
+++ b/conduit/data/datamodules/vision/camelyon17.py
@@ -1,4 +1,4 @@
-"""Waterbirds data-module."""
+"""Camelyon17 data-module."""
 from typing import Any
 
 import albumentations as A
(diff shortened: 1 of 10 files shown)


conduit/data/datamodules/vision/camelyon17.py
```diff
@@ -0,0 +1,78 @@
...
+    Camelyon17SplitScheme,
+)
+from conduit.data.structures import TrainValTestSplit
+
+__all__ = ["Camelyon17DataModule"]
+
+
+@attr.define(kw_only=True)
+class Camelyon17DataModule(CdtVisionDataModule):
+    """Data-module for the Waterbirds dataset."""
```
olliethomas, line 24: ^

conduit/data/datasets/vision/camelyon17.py
```diff
@@ -0,0 +1,170 @@
...
+    For the mixed-to-test setting, slide 23 (corresponding to patient 042, node 3 in the
+    original dataset) is moved from the test set to the training set
+    """
+
+
+class Camelyon17Split(Enum):
+    train = 0
+    id_val = 1
+    test = 2
+    val = 3
```
olliethomas, line 36: can this be `auto()`

conduit/data/datasets/vision/camelyon17.py
```diff
@@ -0,0 +1,170 @@
...
+
+        if self.split is not None:
+            split_indices = self.metadata["split"] == self.split.value
+            self.metadata = cast(pd.DataFrame, self.metadata[split_indices])
+
+        # Construct filepaths from metadata
+        def build_fp(row: pd.DataFrame) -> str:
+            return "patches/patient_{0}_node_{1}/patch_patient_{0}_node_{1}_x_{2}_y_{3}.png".format(
+                *row
+            )
```
olliethomas, line 158: I don't love this, but I can offer no better solution

conduit/data/datasets/vision/camelyon17.py
MylesBartlett: It could be but when the values of the enum actually have some meaning I think it's better to be safe and not rely upon the ordering to auto-generate them.

conduit/data/datasets/vision/camelyon17.py
MylesBartlett: Believe me, this is far nice than the original implementation (derived from WILDS) which made use of ``iiter_tuples``

PR #126, opened 2022-03-15
Implement stratified train-test split with specifiable train-set proportions.

MylesBartlett:


conduit/data/datasets/vision/nico.py
```diff
@@ -131,8 +130,8 @@ def train_test_split(
         *,
         train_props: Optional[Dict[Union[str, int], Dict[Union[str, int], float]]] = None,
         seed: Optional[int] = None,
-    ) -> NICOTrainTestSplit:
-        """Split the data into train/test sets with the option to condition on concept/context."""
+    ) -> TrainTestSplit["NICO"]:
```
olliethomas, line 133:
```suggestion
    ) -> TrainTestSplit[NICO]:
```

conduit/data/datasets/base.py
```diff
@@ -200,9 +199,7 @@ def random_split(self: D, props: Union[Sequence[float], float], deep: bool = Fal
 
         :returns: Random subsets of the data of the requested proportions.
         """
-        # lazily import make_subset to prevent it being a circular import
-        from conduit.data.datasets.utils import make_subset
+        # lazily import ``random_split`` to prevent it being a circular import
+        from conduit.data.datasets.utils import random_split
```
olliethomas, line 203: isn't this solvable by moving where random_split lives?

conduit/data/datasets/vision/nico.py
MylesBartlett: This doesn't work since we're not using ``annotations``.

PR #128, opened 2022-03-18
Proof of concept for nested enum

thomkeh:
Neoconfigen ran without error, but it produced an incorrect import. I had to fix that manually.

```
Python 3.9.7 (default, Sep 10 2021, 14:59:43)
[GCC 11.2.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from conduit.hydra.conduit.data.datamodules.conf import NICODataModuleConf
>>> c = NICODataModuleConf()
>>> c.superclass
Superclass.animals
>>> type(c.superclass)
<enum 'Superclass'>
```


olliethomas:
have you tried with hydra?

thomkeh:
Do you have a test case?

PR #129, opened 2022-03-18
Obviously much better way to do nested enums

thomkeh:
I ran `generate_confs.sh` on it and it produced the same output as before.

### DIFF
--- a/conduit/data/datamodules/vision/nico.py
+++ b/conduit/data/datamodules/vision/nico.py
@@ -9,7 +9,7 @@ from ranzen import implements
 from conduit.data.datamodules.base import CdtDataModule
 from conduit.data.datamodules.vision.base import CdtVisionDataModule
 from conduit.data.datasets.utils import stratified_split
-from conduit.data.datasets.vision.nico import NICO, NicoSuperclass
+from conduit.data.datasets.vision.nico import NICO
 from conduit.data.structures import TrainValTestSplit
 
 __all__ = ["NICODataModule"]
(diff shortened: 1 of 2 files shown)


thomkeh:
Btw, how often do you run the `--runslow` and `--rungpu` tests?

olliethomas:
> Btw, how often do you run the `--runslow` and `--rungpu` tests?

Not very often. I only run them when I'm convinced something is broken. Also, the `--rungpu` flag triggers both run slow and run gpu. I don't know how to make that not be the case

MylesBartlett (approved):

thomkeh:
I have discovered a problem with this: it doesn't work for `dataclasses`/`attrs` classes. At least not if you want to use the `TypeAlias` annotation, which I think we do.

thomkeh:
There is a workaround, but it's not very nice. The trick is to not use `TypeAlias` and to wrap the Enum type in `Union`:

```python
from dataclasses import dataclass
from enum import Enum, auto
from typing import Union

class ASplit(Enum):
    x = auto()
    y = auto()

@dataclass
class A:
    Split = Union[ASplit]

    split: Split = Split.x

a = A(split=A.Split.y)
```
but pyright is actually not completely happy with this.

thomkeh:
Miraculously, the following solution satisfies both mypy and pyright:

```python
from dataclasses import dataclass
from enum import Enum, auto
from typing import TYPE_CHECKING, TypeAlias

class LongEnumName(Enum):
    x = auto()
    y = auto()

@dataclass
class A:
    if TYPE_CHECKING:
        ShortName: TypeAlias = LongEnumName
    else:
        ShortName = LongEnumName

    var: ShortName = ShortName.x


a = A(A.ShortName.y)
print(a.var)
```

PR #131, opened 2022-03-22
Moar Generics

MylesBartlett:
- Add to_device, astuple, and asddict methods to SampleBase.
- Allow device to be optional.
- Make SampleBase generic in x.
- .
- .
- Make datasets generic in s and y.

### DIFF
--- a/conduit/data/datasets/audio/base.py
+++ b/conduit/data/datasets/audio/base.py
@@ -8,7 +8,7 @@ import torch
 from torch import Tensor
 import torchaudio  # type: ignore
 
-from conduit.data.datasets.base import CdtDataset
+from conduit.data.datasets.base import CdtDataset, I, S, Y
 from conduit.data.datasets.utils import (
     AudioLoadingBackend,
     AudioTform,
@@ -20,7 +20,7 @@ from conduit.data.structures import IndexType, TargetData
 __all__ = ["CdtAudioDataset"]
 
 
-class CdtAudioDataset(CdtDataset):
+class CdtAudioDataset(CdtDataset[I, npt.NDArray[np.string_], Y, S]):
     """Base dataset for audio data."""
 
     x: npt.NDArray[np.string_]
--- a/conduit/data/datasets/base.py
+++ b/conduit/data/datasets/base.py
@@ -62,9 +62,9 @@ class CdtDataset(DatasetProt[I], Generic[I, X, Y, S]):
         self, *, x: X, y: Optional[TargetData] = None, s: Optional[TargetData] = None
     ) -> None:
         self.x = x
-        if not isinstance(y, Tensor):
+        if isinstance(y, np.ndarray):
             y = torch.as_tensor(y)
-        if not isinstance(s, Tensor):
+        if isinstance(s, np.ndarray):
             s = torch.as_tensor(s)
         self.y: S = y if y is None else y.squeeze()
         self.s: S = s if s is None else s.squeeze()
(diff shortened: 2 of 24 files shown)


conduit/data/datasets/base.py
```diff
@@ -21,23 +32,42 @@
...
+        self, *, x: X, y: Optional[TargetData] = None, s: Optional[TargetData] = None
     ) -> None:
         self.x = x
         if isinstance(y, np.ndarray):
             y = torch.as_tensor(y)
         if isinstance(s, np.ndarray):
             s = torch.as_tensor(s)
-        self.y = y if y is None else y.squeeze()
-        self.s = s if s is None else s.squeeze()
+        self.y: S = y if y is None else y.squeeze()
```
thomkeh, line 69:
```suggestion
        self.y: Y = y if y is None else y.squeeze()
```

conduit/data/datasets/base.py
```diff
@@ -21,23 +32,42 @@
...
 
 
-class CdtDataset(DatasetProt[NamedSample]):
+class CdtDataset(DatasetProt[I], Generic[I, X, Y, S]):
     _repr_indent: ClassVar[int] = 4
     _logger: Optional[logging.Logger] = None
 
     def __init__(
-        self, *, x: UnloadedData, y: Optional[TargetData] = None, s: Optional[TargetData] = None
+        self, *, x: X, y: Optional[TargetData] = None, s: Optional[TargetData] = None
```
thomkeh, line 62: I assume there is a reason why this is not
```suggestion
        self, *, x: X, y: Y = None, s: S = None
```
?

conduit/data/datasets/base.py
```diff
@@ -62,21 +92,28 @@ def logger(self) -> logging.Logger:
...
+    @overload
+    def _sample_x(self, index: IndexType, *, coerce_to_tensor: Literal[False] = ...) -> LoadedData:
+        ...
+
     def _sample_x(
-        self, index: Union[int, List[int], slice], *, coerce_to_tensor: bool = False
-    ) -> Tensor:
-        self.x
+        self, index: IndexType, *, coerce_to_tensor: bool = False
+    ) -> Union[LoadedData, Tensor]:
```
thomkeh, line 105: Why is the return type not
```suggestion
    ) -> X:
```

conduit/data/datasets/audio/base.py
```diff
@@ -20,7 +20,7 @@
 __all__ = ["CdtAudioDataset"]
 
 
-class CdtAudioDataset(CdtDataset):
+class CdtAudioDataset(CdtDataset[I, npt.NDArray[np.string_], Y, S]):
```
thomkeh, line 23: It seems weird to me that no method signature in this class makes use of the typevars, but maybe that's correct.

conduit/data/datasets/base.py
olliethomas: Is this because `Y = Optional[Tensor]`, not `Y = Optional[TargetData]`?

MylesBartlett: This has to do with the fact that numpy arrays are converted into tensors in the init

conduit/data/datasets/base.py
thomkeh: Did you address this?

MylesBartlett: I committed your suggestion (or tried to)

MylesBartlett, line 69:
```suggestion
        self.y: Y = y if y is None else y.squeeze()
```

thomkeh:
Sorry, my PR caused merge conflicts.

PR #132, opened 2022-03-22
Clarify that Python 3.8 is the minimal required version

thomkeh:


MylesBartlett (approved):

thomkeh:
hm, the CI had been silently using a different version of python than specified

image

PR #133, opened 2022-03-27
Add group-wise metrics.

MylesBartlett:


olliethomas (approved):
lgtm

conduit/metrics.py
```diff
@@ -0,0 +1,202 @@
...
+    "Aggregate by taking the minimum."
+    MAX = max
+    "Aggregate by taking the maximum."
+    MEAN = torch.mean
+    "Aggregate by taking the mean."
+    MEADIAN = torch.median
+    "Aggregate by taking the median."
+    DIFF = partial(_pdist_1d)
+    "Aggregate by taking the pairwise (absolute) differences."
+    MAX_DIFF = partial(_max_diff_1d)
```
thomkeh, line 94: Why are these partials? Is it because otherwise the functions become methods or something like that?

conduit/models/erm.py
```diff
@@ -85,7 +80,7 @@ def inference_epoch_end(self, outputs: EPOCH_OUTPUT, stage: Stage) -> MetricDict
             acc1, acc5 = precision_at_k(logits=logits_all, targets=targets_all, top_k=(1, 5))
```
thomkeh, line 80: You removed the import of `precision_at_k` so I don't understand how this still works

conduit/metrics.py
```diff
@@ -0,0 +1,202 @@
...
+        *,
+        y_true: Tensor,
+        s: Tensor,
+    ) -> Tensor:
+        """
+        Compute the subclass-wise score(s) using the given comparator function.
+
+        :param y_pred: Predicted labels or raw logits of a classifier.
+        :param y_true: Ground truth (correct) labels.
+        :param s: Ground turth labels indicating the subclass-membership of each sample.
```
thomkeh, line 131:
```suggestion
        :param s: Ground truth labels indicating the subclass-membership of each sample.
```

conduit/metrics.py
```diff
@@ -0,0 +1,202 @@
...
+    "Aggregate by taking the pairwise (absolute) differences."
+    MAX_DIFF = partial(_max_diff_1d)
+    "Aggregate by taking the maximum of the pairwise (absolute) differences."
+
+
+C = TypeVar("C", bound=Comparator, covariant=True)
+
+
+class per_subclass_metric(Generic[C]):
+    def __init__(self, comparator: C, aggregator: Optional[Union[Aggregator, str]] = None) -> None:
```
thomkeh, line 102: The genericness seems to only be used in the constructor, making it somewhat pointless.

thomkeh:
Does this replace the EthicML metrics?

conduit/models/erm.py
MylesBartlett: Ah, it wouldn't do if it were actually run -- the only models that are tested are the fair ones. As I said, I hope to remove the models from conduit at some point and lay them to rest in a model graveyard repo so I won't be shoring this up.

conduit/metrics.py
MylesBartlett: I have no idea why the partials aren't required for my own functions but not for the built-in and torch ones, but alas that's how it is (specifically without partial application the values seemingly become the names, bafflingly)

MylesBartlett:
> Does this replace the EthicML metrics?

Maybe for torch-based code as these functions don't require one to convert to numpy arrays, ``Prediction`` and data-tuples before invocation (while also being being faster due to vectorisation of the subclass-wise ops)

conduit/metrics.py
MylesBartlett: The genericness allows the arguments of the comparator to be propagated. 
To demonstrate, with the genericness pyright shows this function signature
``robust_accuracy: per_subclass_metric[(y_pred: Tensor, *, y_true: Tensor)``
and without it, this one
``robust_accuracy: per_subclass_metric``

MylesBartlett:
I think I'm going to try to unify this code further, superseding ``conditional_equals`` with something like a ``condition_on_y`` argument

conduit/metrics.py
thomkeh: You could use the trick I used in EthicML which is to use a 1-tuple to wrap the callables.

MylesBartlett: That works?

MylesBartlett: Ah but you do need to unpack if you do it that way

MylesBartlett: with partial, on the other hand, I can simply call the value which means there's no need for special-casing

thomkeh: Well, you would wrap all the enum values, so when you call the functions, you'd do `e.value[0](args)`. But I can understand if you prefer the `partial` solution

thomkeh: It's a general problem with assigning functions to variables in class bodies, not just Enum: https://stackoverflow.com/a/28415970

conduit/metrics.py
```diff
@@ -0,0 +1,249 @@
...
+    for k in top_k:
+        correct_k = correct[:k].contiguous().view(-1).float().sum(0, keepdim=True)
+        res.append(correct_k.mul_(100.0 / batch_size))
+    return res
+
+
+R = TypeVar("R", Tensor, Tuple[Tensor, Union[Tensor, slice]], covariant=True)
+
+
+@runtime_checkable
```
thomkeh, line 65: There is little point in making this runtime checkable because it will just check whether the object is callable.

PR #134, opened 2022-04-04
input container

MylesBartlett:
- Expand LoadedData to include dicts.
- General _add_xs.

### DIFF
--- a/conduit/data/datasets/vision/base.py
+++ b/conduit/data/datasets/vision/base.py
@@ -1,3 +1,5 @@
+from functools import reduce
+import operator
 from pathlib import Path
 from typing import List, Optional, Sequence, Union, cast, overload
 
@@ -16,10 +18,10 @@ from conduit.data.datasets.utils import (
     apply_image_transform,
     img_to_tensor,
     infer_il_backend,
-    is_tensor_list,
     load_image,
 )
 from conduit.data.structures import IndexType, TargetData
+from conduit.types import Addable
 
 __all__ = ["CdtVisionDataset"]
 
@@ -82,8 +84,12 @@ class CdtVisionDataset(CdtDataset[I, npt.NDArray[np.string_], Y, S]):
             index = list(range(len(self)))[index]
         if isinstance(index, list):
             sample_ls = [self._sample_x(index=i, coerce_to_tensor=coerce_to_tensor) for i in index]
-
-            if is_tensor_list(sample_ls):
+            elem = sample_ls[0]
+            if isinstance(elem, Addable):
+                summed = reduce(operator.add, sample_ls)
+                return cast(ItemType, summed)
+            if isinstance(elem, Tensor):
+                sample_ls = cast(List[Tensor], sample_ls)
                 return torch.stack(sample_ls, dim=0)
             elif isinstance(sample_ls[0], np.ndarray):
                 return np.stack(sample_ls, axis=0)
--- a/conduit/data/structures.py
+++ b/conduit/data/structures.py
@@ -15,7 +15,6 @@ from typing import (
     Union,
     cast,
     overload,
-    runtime_checkable,
 )
 
 from PIL import Image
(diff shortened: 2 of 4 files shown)


conduit/data/datasets/vision/base.py
```diff
@@ -82,8 +84,12 @@ def _sample_x(
             index = list(range(len(self)))[index]
         if isinstance(index, list):
             sample_ls = [self._sample_x(index=i, coerce_to_tensor=coerce_to_tensor) for i in index]
-
-            if is_tensor_list(sample_ls):
+            elem = sample_ls[0]
+            if isinstance(elem, Addable):
+                summed = reduce(operator.add, sample_ls)
```
thomkeh, line 89: Why not use `sum()`?

conduit/data/structures.py
```diff
@@ -49,13 +51,90 @@
...
+        for key, value in x2.items():  # type: ignore
+            if key in x1:
+                x1[key] = concatenate_inputs(x1[key], value, is_batched=is_batched)  # type: ignore
+            else:
+                x1[key] = value  # type: ignore
+            return x1
+    return x1 + x2  # type: ignore
+
+
+class InputContainer(Sized[X_co], Addable):
```
thomkeh, line 125: Shouldn't this be a protocol as well?

conduit/data/structures.py
```diff
@@ -76,97 +155,47 @@ def shape(self):
...
-TargetData: TypeAlias = Union[Tensor, npt.NDArray[np.floating], npt.NDArray[np.integer]]
-
-IndexType: TypeAlias = Union[int, List[int], slice]
-
-
-X = TypeVar("X", bound=LoadedData)
-
-
-@dataclass
+@dataclass(init=False)
```
thomkeh, line 176: What's the point of a dataclass without init?

conduit/types.py
```diff
@@ -42,3 +44,17 @@ class Stage(Enum):
...
 MetricDict: TypeAlias = Dict[str, _METRIC_COLLECTION]
 NDArrayR: TypeAlias = Union[npt.NDArray[np.floating], npt.NDArray[np.integer]]
+
+T_co = TypeVar("T_co", covariant=True)
+
+
+@runtime_checkable
+class Sized(Protocol[T_co]):
+    def __len__(self) -> int:
+        ...
```
thomkeh, line 54: The genericness seems pointless?

conduit/types.py
```diff
@@ -42,3 +44,17 @@ class Stage(Enum):
...
+
+
+@runtime_checkable
+class Addable(Protocol[T_co]):
+    def __add__(self, other: Self) -> Self:
+        ...
```
thomkeh, line 60: Same here

conduit/data/datasets/vision/base.py
MylesBartlett: Because sum initialises the summand as 0 and thus requires any class implementing ``__add__`` to be addable to integers.

conduit/data/structures.py
MylesBartlett: Could be

conduit/types.py
MylesBartlett: It's not pointless if you want to make use of genericness in the child classes

conduit/data/datasets/vision/base.py
thomkeh: Oh, I didn't know.

conduit/types.py
thomkeh: But then just make the child class generic.

MylesBartlett: That's an option but is there a downside to doing it this way?

MylesBartlett: To me, there seems little reason not to parameterise a protocol and at least give the option to capitalise on it downstream

conduit/data/datasets/vision/base.py
thomkeh: It seems you could do this:

```python
sum(a[1:], a[0])
```

MylesBartlett: Yeah, that works but summing over a list doesn't -- I experimented with this quite a bit before stumbling on the reduce+operator solution.

PR #138, opened 2022-04-20
Audio

olliethomas:


conduit/data/datasets/audio/ecoacoustics.py
```diff
@@ -137,15 +123,16 @@ def __init__(
...
 
     @property
-    def segment_len(self) -> Optional[float]:
+    def segment_len(self) -> float:
         return self._segment_len
 
     @segment_len.setter
-    def segment_len(self, value: Optional[float]) -> None:
+    def segment_len(self, value: float) -> None:
+        assert (
```
MylesBartlett, line 131: raise a Value/Attribute error instead of asserting

MylesBartlett: That said, rather than raising an error when the upper bound is violated, it might be better to simply cap the segment length.

MylesBartlett (approved):

conduit/data/datasets/audio/ecoacoustics.py
```diff
@@ -74,10 +74,8 @@ class Ecoacoustics(CdtAudioDataset[SampleType, Tensor, Tensor]):
         ),
     ]
 
-    SAMPLE_RATE: ClassVar[int] = 48_000
-    AUDIO_LEN: ClassVar[int] = 60
-    NUM_FRAMES_TOTAL: ClassVar[int] = SAMPLE_RATE * AUDIO_LEN
     _num_frames_in_segment: Optional[int]
+    _MAX_AUDIO_LEN = 60
```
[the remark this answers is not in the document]
olliethomas: maybe iterating through the dataset once and setting the max length to the min available length? Or getting the mode value and rejecting any samples less than this?

PR #140, opened 2022-04-21
Fix problems encountered when using audio data

olliethomas:
when I've actually used the dataset, there's a few things that didn't work.

That said, there's still a lot more to do. eg default transforms etc

### DIFF
--- a/conduit/transforms/audio.py
+++ b/conduit/transforms/audio.py
@@ -8,7 +8,7 @@ import torchaudio.transforms
 
 Mels = torchaudio.transforms.MelSpectrogram
 
-__all__ = ["LogMelSpectrogram", "LogMelSpectrogramNp", "Framing", "FramingNp"]
+__all__ = ["LogMelSpectrogram", "Framing"]
 
 
 class LogMelSpectrogram(torchaudio.transforms.MelSpectrogram):
--- a/conduit/transforms/audio.py
+++ b/conduit/transforms/audio.py
@@ -1,0 +1,270 @@
+from typing import Union
+
+import numpy as np
+import torch
+from torch import nn
+import torchaudio.transforms
+
+Mels = torchaudio.transforms.MelSpectrogram
+
+__all__ = ["LogMelSpectrogram", "Framing"]
+
+
+class LogMelSpectrogram(nn.Module):
+    # Mel spectrum constants and functions.
+
+    def __init__(
+        self,
+        sample_rate: int = 16000,
+        stft_window_length_seconds: float = 0.025,
+        stft_hop_length_seconds: float = 0.010,
+        log_offset: float = 0.01,  # Offset used for stabilized log of input mel-spectrogram.
+        num_mel_bins: int = 64,  # Frequency bands in input mel-spectrogram patch.
+        mel_min_hz: float = 125,
+        mel_max_hz: float = 7500,
+        mel_break_freq_hertz: float = 700.0,
+        mel_high_freq_q: float = 1127.0,
+    ):
+        super().__init__()
+        self.window_length_samples = int(round(sample_rate * stft_window_length_seconds))
+        self.hop_length_samples = int(round(sample_rate * stft_hop_length_seconds))
+        self.fft_length = 2 ** int(np.ceil(np.log(self.window_length_samples) / np.log(2.0)))
+        self.log_offset = log_offset
+        self.audio_sample_rate = sample_rate
+        self.num_mel_bins = num_mel_bins
+        self.lower_edge_hertz = mel_min_hz
+        self.upper_edge_hertz = mel_max_hz
+        self.mel_break_freq_hertz = mel_break_freq_hertz
+        self.mel_high_freq_q = mel_high_freq_q
+
+    def forward(self, data: torch.Tensor) -> torch.Tensor:
+        """Convert waveform to a log magnitude mel-frequency spectrogram.
+        Args:
+          data: 1D np.array of waveform data.
+          audio_sample_rate: The sampling rate of data.
+          log_offset: Add this to values when taking log to avoid -Infs.
+          window_length_secs: Duration of each window to analyze.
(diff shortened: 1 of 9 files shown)


conduit/transforms/audio.py
```diff
@@ -0,0 +1,352 @@
...
+        _, _, num_samples = data.shape
+        num_frames = 1 + int(np.floor((num_samples - window_length) / hop_length))
+        strides = (1, 1, hop_length)
+
+        return (
+            data.unsqueeze(1)
+            .as_strided(size=(num_frames, 64, window_length), stride=strides)
+            .unsqueeze(1)
+            .permute(0, 1, 3, 2)
+        )
```
olliethomas, line 305: I am not at all confident about this

olliethomas:
This is WIP. The thing that I need to sort next is the collation. I'm not sure of the best way to do this

conduit/transforms/audio.py
MylesBartlett: You can just use ```movedim`` here as you're only permuting two dimensions

MylesBartlett: That aside, I have no idea as to the correctness of this snippet

olliethomas:
Added a converter function to be passed to the `collate_fn`. Everything seems to work. Except of course for the torch transforms that run, but I'm pretty sure aren't right

conduit/transforms/audio.py
```diff
@@ -0,0 +1,352 @@
...
+            spectrogram,
+            self.spectrogram_to_mel_matrix(
+                num_spectrogram_bins=spectrogram.shape[1],
+                audio_sample_rate=self.audio_sample_rate,
+                num_mel_bins=self.num_mel_bins,
+                lower_edge_hertz=self.lower_edge_hertz,
+                upper_edge_hertz=self.upper_edge_hertz,
+            ),
+        )
+        return torch.Tensor(np.log(mel_spectrogram + self.log_offset), device=data.device)
```
thomkeh, line 123: `torch.from_numpy`

thomkeh (commented):
So, why did you keep the numpy version?

conduit/transforms/audio.py
```diff
@@ -0,0 +1,352 @@
...
+          audio_sample_rate: The sampling rate of data.
+          log_offset: Add this to values when taking log to avoid -Infs.
+          window_length_secs: Duration of each window to analyze.
+          hop_length_secs: Advance between successive analysis windows.
+          **kwargs: Additional arguments to pass to spectrogram_to_mel_matrix.
+        Returns:
+          2D np.array of (num_frames, num_mel_bins) consisting of log mel filterbank
+          magnitudes for successive frames.
+        """
+        np_data = data.numpy()[0]
```
thomkeh, line 106: Why do you only use the first element of the batch?

MylesBartlett: Because he's basic

MylesBartlett: No, this is what he's referring to when he says the 'collation problem'

conduit/transforms/audio.py
olliethomas: seems like the sort of thing that should be consistent

conduit/transforms/audio.py
olliethomas: this isn't the collation problem. This is because the torch tensor is of shape `(N,1)` and the numpy equivalent is shape `(N)`

olliethomas: numpy equivalent being if I loaded the data using soundfile and wrapped in numpy

MylesBartlett: ah, then wouldn't it be better to flatten?

thomkeh: But then it should be `[:, 0]`, no?

thomkeh: Or maybe it's always a single sample

thomkeh: It's safer to use `squeeze` anyway

conduit/transforms/audio.py
thomkeh: `as_tensor` then

olliethomas:
> So, why did you keep the numpy version?

because the torch version produces _an_ output, but I'm not confident in it at all. So the plan is to use the numpy one for the pretrained model, and maybe use the torch one for future training.

conduit/transforms/audio.py
olliethomas: transforms are called on a single sample

thomkeh, line 106:
```suggestion
        np_data = data.numpy().squeeze(axis=0)
```

MylesBartlett: but they don't have to be

thomkeh: This is safer because it will tell you if you were mistaken about the assumption that the first dimension is 1

thomkeh:
I'm pretty confident in the torch output actually. What makes you suspicious?

conduit/transforms/audio.py
```diff
@@ -0,0 +1,352 @@
...
+            wkwargs=wkwargs,
+            center=center,
+            pad_mode=pad_mode,
+            onesided=onesided,
+            norm=norm,
+            mel_scale=mel_scale,
+        )
+        self.log_offset = log_offset
+
+    def forward(self, waveform: torch.Tensor) -> torch.Tensor:
```
MylesBartlett, line 61:
```suggestion
    def forward(self, waveform: torch.Tensor) -> torch.Tensor:
```
Can Tensor from torch (rather than accessed through it) for consistency?

olliethomas:
> I'm pretty confident in the torch output actually. What makes you suspicious?

the actual values are pretty different, although not wildly, and the `as_strided` I feel I hacked together a bit

MylesBartlett (approved):

PR #142, opened 2022-05-04
Fix things.

MylesBartlett:
Fix concatenation of s/y in sample classes' __add__ method.

Fix audio datamodules.

### DIFF
--- a/conduit/data/datamodules/audio/base.py
+++ b/conduit/data/datamodules/audio/base.py
@@ -4,7 +4,7 @@ from typing import Optional, TypeVar
 import albumentations as A  # type: ignore
 import attr
 from ranzen.decorators import implements
-import torchaudio.transforms
+import torchaudio.transforms as T  # type: ignore
 from typing_extensions import final
 
 from conduit.data.datamodules.base import CdtDataModule
@@ -60,14 +60,12 @@ class CdtAudioDataModule(CdtDataModule[D, I]):
             self._test_data.transform = transform
 
     @property
-    def _default_train_transforms(self) -> A.Compose:
-        transform_ls: List[AlbumentationsTform] = [torchaudio.transforms.Spectrogram()]
-        return A.Compose(transform_ls)
+    def _default_train_transforms(self) -> T.Spectrogram:
+        return T.Spectrogram()
 
     @property
-    def _default_test_transforms(self) -> A.Compose:
-        transform_ls: List[AlbumentationsTform] = [torchaudio.transforms.Spectrogram()]
-        return A.Compose(transform_ls)
+    def _default_test_transforms(self) -> T.Spectrogram:
+        return T.Spectrogram()
 
     @implements(CdtDataModule)
     @final
--- a/conduit/data/datamodules/audio/ecoacoustics.py
+++ b/conduit/data/datamodules/audio/ecoacoustics.py
@@ -1,21 +1,19 @@
 """Ecoacoustics data-module."""
-from typing import Any, List, Optional, Sequence
+from typing import Any, List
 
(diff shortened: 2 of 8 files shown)


conduit/data/datamodules/audio/base.py
```diff
@@ -60,14 +60,12 @@ def test_transforms(self, transform: Optional[AudioTform]) -> None:  # type: ign
             self._test_data.transform = transform
 
     @property
-    def _default_train_transforms(self) -> A.Compose:
-        transform_ls: List[AlbumentationsTform] = [torchaudio.transforms.Spectrogram()]
-        return A.Compose(transform_ls)
+    def _default_train_transforms(self) -> T.Spectrogram:
```
olliethomas, line 63: hmmm, should the return type be `nn.Module`?

conduit/data/structures.py
```diff
@@ -612,16 +599,18 @@ def __len__(self) -> int:
 
 
 @runtime_checkable
-class DatasetWrapper(SizedDataset[R_co], Protocol[D]):
-    dataset: D
+class DatasetWrapper(SizedDataset[R_co]):
```
olliethomas, line 602: gives 
```
E   TypeError: @runtime_checkable can be only applied to protocol classes, got <class 'conduit.data.structures.DatasetWrapper'>
```

olliethomas:
> Can't put this in the review as it is on an unchanged line.
> 
> ```
>         num_frames_segment = int(
> >           self.num_frames_in_segment / self.sample_rate * metadata.sample_rate
>         )
> E       AttributeError: 'Ecoacoustics' object has no attribute 'num_frames_in_segment'
> 
> ../conduit/data/datasets/audio/ecoacoustics.py:266: AttributeError
> ```

i think lines 266 and 270 just need to have an underscore at the start of the num_frame_in_segment var

MylesBartlett:
> > Can't put this in the review as it is on an unchanged line.
> > ```
> >         num_frames_segment = int(
> > >           self.num_frames_in_segment / self.sample_rate * metadata.sample_rate
> >         )
> > E       AttributeError: 'Ecoacoustics' object has no attribute 'num_frames_in_segment'
> > 
> > ../conduit/data/datasets/audio/ecoacoustics.py:266: AttributeError
> > ```
> 
> i think lines 266 and 270 just need to have an underscore at the start of the num_frame_in_segment var

No the problem was with the setter using the underscored version.

conduit/data/datamodules/audio/base.py
MylesBartlett: No, it could be AudioTForm but prefer narrower return types.

olliethomas:
If line 154 of tests/data_test.py is removed so that the default transforms are used, there is an error:
```
        # Test size().
>       assert test_sample.x.size()[1] == dm.dims[0]
E       assert 1 == 75

data_test.py:164: AssertionError
```

conduit/data/datamodules/audio/ecoacoustics.py
```diff
@@ -28,29 +26,7 @@ class EcoacousticsDataModule(CdtAudioDataModule[Ecoacoustics, TernarySample]):
...
-            ds,
-            batch_size=batch_size if batch_sampler is None else 1,
-            shuffle=shuffle,
-            num_workers=self.num_workers,
-            pin_memory=self.pin_memory,
-            drop_last=drop_last,
-            persistent_workers=self.persist_workers,
-            batch_sampler=batch_sampler,
-            converter=self.converter,
-        )
```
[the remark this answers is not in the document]
MylesBartlett: Ah you're right; I took this for duplicate code

PR #148, opened 2022-06-06
Fix: allow [m, n] labels over [1, n] labels

m4gpi:
Removes reshaping the y dimensions along the same axis as x (effectively cloning the labels). This means we can now support `[m, n]` shaped labels yielding multiple categories for labels, rather than just `[n]` shaped labels.

Uncertain about whether to reshape `y` such that when there is only one dimension it's shape is `[1, n]` rather than `[n]` (i.e. a vector rather than a 1D array). Looked at using `torch.atleast_2d` but returns values extended along the row rather than column (opposite of when labels are already 2d). No way to know in advance dimensions should be transposed or not.


MylesBartlett:
This behaviour of the ```atleast``` functions is a continual gripe of mine -- they'd be far more useful imo if it expanded rightwards. Anyway, I think you should `squeeze` the labels such that when they're 1-dimensional they're simply of shape ```[N,]```.  The batch dimension should always come first, by the way, such that ```y`` is of shape ```[N, M]``` when ```M > 1``.

tests/data_test.py
```diff
@@ -166,6 +166,23 @@ def test_ecoacoustics_dm(root: Path):
     assert test_sample.x.size()[3] == dm.dims[2]
 
 
+@pytest.mark.slow
+def test_ecoacoustics_dm_batch_multi_label(root: Path) -> None:
```
olliethomas, line 170: Check that this PR works with varying batch_sizes. You can parametrize the test with the below, and the pass this as an arg to the dataloader

```suggestion
@pytest.mark.slow
@pytest.mark.parametrize("batch_size", [1, 2, 8,])
def test_ecoacoustics_dm_batch_multi_label(root: Path, batch_size: int) -> None:
```

MylesBartlett:
At some point we might want to make it so that ```y``` and ```s``` can take forms other than tensors so we can rely on keys/attributes instead of ordering for selecting the different targets.

thomkeh:
Does the new test pass? (It's not run by the CI because it's marked as slow.)

olliethomas:
> Does the new test pass? (It's not run by the CI because it's marked as slow.)

I've just checked out the branch and run both tests. They pass

PR #149, opened 2022-06-06
Remove models and architectures in preparation for pypi publication.

MylesBartlett:
Models have been moved to a separate repo, https://github.com/wearepal/legacy_models, in light of the fact that most (all?) are broken or unused and pre-defining models is generally not very useful as the callbacks need to be tailored to the task in question.


MylesBartlett:
> You might want to move `conduit/models/utils.py` somewhere else because it is now the only file in `conduit/models`.

I've given great thought to this but decided to leave it as it is for now so as not to break the repos which import from conduit.models.utils.

PR #152, opened 2022-06-06
Use new reduce_scatter function for groupwise metrics instead of broadcasting.

MylesBartlett:
- Use newly-added reduce_scatter function for computing groupwise metrics instead of broadcasting.
- Compute index set using inverse indices.


thomkeh (approved):
Fine apart from the question about torch version

MylesBartlett:
Currently running ```poetry update``` :snorlax:

PR #154, opened 2022-06-07
Release script: auto-merge the PR

thomkeh:
I don't know why I didn't think of this earlier.


thomkeh:
Hm, PRs here still require a review...

olliethomas:
I think that this is because of the branch protections. We can bypass them from a specific user e.g. Dependabot isn't subject to approval. The problem is we'd need a specific user to be exempt. We could use a machine-user for this, but we need a server ideally (🤞this happens soon)

PR #155, opened 2022-06-08
bug/fix label mapping

m4gpi:
some attributes in the `metadata.csv` file are capitalised or partially so (e.g. `NN, N0, H.s`). These were not so in the `SoundscapeAttr` enum definition, and thus `enum.name` did not match the field in the metadata file, causing an error when attempting to load these attributes. This capitalises the enum names so they match the fixed ones in the metadata file. 

Second issue, when parsing provided attributes to the dataset (either as string or enum), the output was being downcased, causing the same error to occur. Fixed by removing `str(...)` which was downcasing the field name.

```
>>> x = [str_to_enum(str_=elem.name, enum=SoundscapeAttr) for elem in target_attrs]
>>> x[3]
SoundscapeAttr.NN
>>> x[3].name
'NN'
>>> str(x[3])
'nn'
```

### DIFF
--- a/conduit/data/datasets/audio/ecoacoustics.py
+++ b/conduit/data/datasets/audio/ecoacoustics.py
@@ -39,8 +39,8 @@ class SoundscapeAttr(Enum):
     habitat = auto()
     site = auto()
     time = auto()
-    nn = auto()
-    n0 = auto()
+    NN = auto()
+    N0 = auto()
 
 
 SampleType: TypeAlias = TernarySample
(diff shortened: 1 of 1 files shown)


thomkeh:
Hm, the fact that `@enum_name_str` calls `.lower()` on the name maybe isn't the best behavior. Would you be okay with changing that @olliethomas @MylesBartlett ?

thomkeh (approved):
Anyway, looks good

PR #158, opened 2022-07-21
add md attributes and switch to using a ternary sample in ecoacoustics

m4gpi:
- removes the use of the batch converter
- switches ecoacoustic dataset to using `TernarySample`
- adds additional attributes enum `MDAttr` which at the moment just contains the `fileName` field from the metadata
- encodes filenames as integers in a tensor
- creates a dictionary decoder to index the filename string
- adds `A_Den` field for fetching `A.Den` target from the metadata (will this work?)

### DIFF
--- a/conduit/data/datasets/audio/ecoacoustics.py
+++ b/conduit/data/datasets/audio/ecoacoustics.py
@@ -9,7 +9,7 @@ from enum import Enum, auto
 import math
 from pathlib import Path
 import shutil
-from typing import ClassVar, List, Optional, Tuple, Union, cast
+from typing import ClassVar, List, Optional, Tuple, Union, cast, Dict
 import zipfile
 
 import numpy as np
@@ -121,10 +121,6 @@ class Ecoacoustics(CdtAudioDataset[SampleType, Tensor, Tensor]):
 
         self.metadata = pd.read_csv(self.base_dir / self._METADATA_FILENAME)
         # map numerical indices to target attributes
-        self.md_decoder = self.metadata[self.md_attrs].to_dict()
-        encoded_targets = [SoundscapeAttr.habitat.name, SoundscapeAttr.site.name]
-        self.target_attr_decoder = self.metadata[encoded_targets].to_dict()
-
         x = self.metadata["filePath"].to_numpy()
         y = torch.as_tensor(
             self._label_encode(self.metadata[self.target_attrs], inplace=True).to_numpy()
@@ -166,6 +162,10 @@ class Ecoacoustics(CdtAudioDataset[SampleType, Tensor, Tensor]):
                 )
             if zipfile.is_zipfile(dir_):
                 raise RuntimeError(f"{dir_} file not unzipped.")
+
+    def label_decoder(self) -> Dict[SoundscapeAttr, Dict[str, int]]:
+        encoded_targets = [SoundscapeAttr.habitat.name, SoundscapeAttr.site.name] + self.md_attrs
+        return self.metadata[encoded_targets].to_dict()
 
     @staticmethod
     def _label_encode(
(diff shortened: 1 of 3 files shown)


conduit/data/datasets/audio/ecoacoustics.py
```diff
@@ -111,13 +120,18 @@ def __init__(
             self._extract_metadata()
 
         self.metadata = pd.read_csv(self.base_dir / self._METADATA_FILENAME)
+        # map numerical indices to target attributes
+        self.concept_label_decoder = self.metadata[self.md_attrs].to_dict()
```
MylesBartlett, line 124:
```suggestion
        self.md_decoder = self.metadata[self.md_attrs].to_dict()
```
Doesn't this need to be paired with the label encoded data?

m4gpi: indeed :joy:

conduit/data/datasets/audio/ecoacoustics.py
```diff
@@ -41,6 +42,12 @@ class SoundscapeAttr(Enum):
     time = auto()
     NN = auto()
     N0 = auto()
+    A_Den = auto()
+
+
+@enum_name_str
+class MDAttr(Enum):
+    fileName = auto()
```
olliethomas, line 50: This isn't really a review, just a comment, that this mix of enum naming styles makes me feel quite anxious.

m4gpi: as in `MD` as opposed to `Md`?

olliethomas: what I mean is that between these two enums, they have members with names
- `habitat`
- `site`
- `time`
- `NN`
- `N0`
- `A_Den`
- `fileName`

In an ideal world there would be some consistency, even though it's not really required for 🐍

m4gpi: that should have been resolved at source, i.e. in the csv files, but you deal with what you have.

conduit/data/datasets/audio/ecoacoustics.py
```diff
@@ -41,6 +42,12 @@ class SoundscapeAttr(Enum):
     time = auto()
     NN = auto()
     N0 = auto()
+    A_Den = auto()
```
MylesBartlett, line 45: doesn't the value need to be manually encoded as "A.Den"?

conduit/data/datasets/audio/ecoacoustics.py
MylesBartlett: I think when the values correspond to column names it's probably best to just manually write the mappings rathr than relying on the decorator (which simply makes str(.) return .name.lower())

m4gpi: so I'll change `fileName` to `file_name`, that will work. I could also change `NN` and `N0` to `species_abundance` and `species_richness`?

conduit/data/datasets/audio/ecoacoustics.py
m4gpi: done

conduit/data/datasets/audio/ecoacoustics.py
m4gpi: not sure if this will have a cascade effect, but I assume not, proposal for naming:

```
@enum_name_str
class SoundscapeAttr(Enum):
    HABITAT = auto()
    SITE = auto()
    TIME = auto()
    SPECIES_ABUNDANCE = "NN"
    SPECIES_RICHNESS = "N0"
    A_DEN = "A.Den"
  
  
@enum_name_str
class MDAttr(Enum):
    FILE_NAME = "fileName"
```

PR #166, opened 2022-08-04
progbar

MylesBartlett:
The lions' share of this code was shamelessly stolen from quaterion
- Add custom progress bar.
- Add rich as a dependency (for progbar).


olliethomas (approved):
Well done

PR #173, opened 2022-08-12
Clone 'seed' ids for index_set in groupwise_metric to prevent in-place modification.

MylesBartlett:


thomkeh:
Shouldn't you actually use the new function?

PR #187, opened 2022-09-22
Set exclude_self in reduce_scatter to False to correct 1-off error.

MylesBartlett:
Use init instead of partial to assign function values to Aggregator enum.

Add more classification metrics and expand metrics test.

### DIFF
--- a/conduit/metrics.py
+++ b/conduit/metrics.py
@@ -90,6 +90,8 @@ def _pdist_1d(x: Tensor) -> Tensor:
 
 
 def max_difference_1d(x: Tensor) -> Tensor:
+    if x.numel() == 1:
+        return x.squeeze()
     return _pdist_1d(x).max()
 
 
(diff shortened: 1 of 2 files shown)


conduit/metrics.py
```diff
@@ -83,27 +86,44 @@ def __call__(self, y_pred: Tensor, *, y_true: Tensor) -> R:
 
 
 def _pdist_1d(x: Tensor) -> Tensor:
-    return torch.pdist(x.view(-1, 1)).squeeze()
+    res = torch.pdist(x.view(-1, 1)).squeeze()
+    return res
```
thomkeh, line 90: ??

thomkeh (approved):

conduit/metrics.py
MylesBartlett: Ah I split up the return for the purpose of testing; the squeeze was causing an assert to fail

PR #32, opened 2021-07-27
add more EM datasets

olliethomas:

PR #209, opened 2022-12-16
Add NIH Chest X-rays dataset.

MylesBartlett:
* Add NIH Chest X-rays dataset (https://www.nih.gov/news-events/news-releases/nih-clinical-center-provides-one-largest-publicly-available-chest-x-ray-datasets-scientific-community)
* Update pyproject.toml to use new group syntax.


conduit/data/datasets/vision/nih.py
```diff
@@ -0,0 +1,186 @@
...
+        Pleural_Thickening    0.030191
+        Pneumonia             0.012763
+        Pneumothorax          0.04728
+    """
+
+    ATELECTASIS = "Atelectasis"
+    CARDIOMEGALY = "Cardiomegaly"
+    CONSOLIDATION = "Consolidation"
+    EDEMA = "Edema"
+    EFFUSION = "Effusion"
```
thomkeh, line 52: are we doing ALL_CAPS enums again?

thomkeh (approved):

conduit/data/datasets/vision/nih.py
MylesBartlett: I think the plan was always to gradually transition to using them throughout conduit as all caps is the convention for enums

MylesBartlett: However, do you think it'd be better to just remain internally consistent?

thomkeh: No, as long as you have a plan, it's fine.

thomkeh:
hm, something about `sklearn` as a dependency?

thomkeh:
EthicML depends on `folktables` which depends on `sklearn` (which should be `scikit-learn`).

thomkeh:
There is an issue for it: https://github.com/zykls/folktables/issues/32

conduit/data/datasets/vision/nih.py
```diff
@@ -0,0 +1,189 @@
...
+    `this
+    <https://www.nih.gov/news-events/news-releases/nih-clinical-center-provides-one-largest-publicly-available-chest-x-ray-datasets-scientific-community>`__
+    Open Access paper.
+
+    The dataset can be downloaded by following the above link or from `kaggle <https://www.kaggle.com/datasets/nih-chest-xrays/data>`__
+    """
+
+    TARGET_ATTR: TypeAlias = NIHTargetAttr
+    SENS_ATTR: TypeAlias = NIHSensAttr
+    SPLIT: TypeAlias = NIHSplit
```
thomkeh, line 82: I don't think the TypeAliases should be in ALL CAPS?

MylesBartlett: 👍

PR #213, opened 2022-12-18
Modernize some enums, fix some other things

thomkeh:
I replaced this pattern
```python
@enum_name_str
class PacsDomain(Enum):
    photo = auto()
    art_painting = auto()
    cartoon = auto()
    sketch = auto()

class X:
    def __init__(self,  domains: PacsDomains | str, ...):
        self.domains = str_to_enum(str_=domains, enum=PacsDomain)
```

with this pattern:
```python
class PacsDomain(StrEnum):
    PHOTO = auto()
    ART_PAINTING = auto()
    CARTOON = auto()
    SKETCH = auto()

class X:
    def __init__(self, domains: PacsDomains | str, ...):
        self.domains = PacsDomain(domains)
```

In terms of the strings these accept, they are completely equivalent. That's because in a `StrEnum`, `auto()` returns the lower case version of the name. so, `PHOTO = auto()` means `PHOTO = "photo"`. Thus, `PacsDomain(domains)` now does what `PacsDomain[domains]` did before.

```
Python 3.11.1 (main, Dec  6 2022, 19:20:21) [GCC 11.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from enum import StrEnum, auto
>>> class PacsDomain(StrEnum):
...     PHOTO = auto()
...     ART_PAINTING = auto()
...     CARTOON = auto()
...     SKETCH = auto()
... 
>>> PacsDomain(PacsDomain.PHOTO)
<PacsDomain.PHOTO: 'photo'>
>>> PacsDomain("photo")
<PacsDomain.PHOTO: 'photo'>
>>> PacsDomain["PHOTO"]
<PacsDomain.PHOTO: 'photo'>
```


MylesBartlett (approved):

thomkeh:
One problem is that this will break code that depends on conduit, because `CelebAttr.Male` was renamed to `CelebAttr.MALE`.

How should we deal with this? @MylesBartlett 

Maybe a 1.0 release?

MylesBartlett:
> One problem is that this will break code that depends on conduit, because `CelebAttr.Male` was renamed to `CelebAttr.MALE`.
> 
> How should we deal with this? @MylesBartlett 
> 
> Maybe a 1.0 release?

Yes, as I said, that's why I wasn't more eager to enact the change. I think a 1.0 release is too much; bumping up the patch version seems more appropriate.

PR #217, opened 2022-12-22
Port iWildCam and fMoW datasets from WILDS

MylesBartlett:


conduit/data/datasets/vision/fmow.py
```diff
@@ -0,0 +1,197 @@
+from enum import auto
+from pathlib import Path
+from typing import ClassVar, Optional, Union
+
+import pandas as pd  # type: ignore
```
thomkeh, line 5: The type ignore shouldn't be necessary anymore because I added the official pandas stubs https://github.com/pandas-dev/pandas-stubs to the dev dependencies.

PR #291, opened 2023-04-27
Add the NICO++ dataset

tmke8:
This is mostly a copy of the NICI code, but with the special
train_props.

### DIFF
--- a/conduit/data/datamodules/vision/nico_plus_plus.py
+++ b/conduit/data/datamodules/vision/nico_plus_plus.py
@@ -1,5 +1,5 @@
 """NICO++ data-module."""
-from typing import Any, Optional
+from typing import Any, List, Optional
 
 import albumentations as A  # type: ignore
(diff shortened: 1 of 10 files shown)


conduit/data/datamodules/vision/nico_plus_plus.py
```diff
@@ -0,0 +1,59 @@
...
+    superclass: NicoPPTarget = NicoPPTarget.AIRPLANE
+    make_biased: bool = True
+
+    @property
+    @override
+    def _default_train_transforms(self) -> A.Compose:
+        base_transforms = A.Compose(
+            [
+                A.Resize(self.image_size, self.image_size),
+                A.CenterCrop(self.image_size, self.image_size),
```
tmke8, line 30: I'm actually a bit confused about this (this is copied from NICO). Is the crop not superfluous?

PR #302, opened 2023-05-11
Split NICO++ according to the paper 'Change is Hard'

tmke8:


conduit/data/datasets/vision/nico_plus_plus.py
```diff
@@ -110,6 +110,8 @@ class NICOPP(CdtVisionDataset[TernarySample, Tensor, Tensor]):
     SampleType: TypeAlias = TernarySample
     Superclass: TypeAlias = NicoPPTarget
     Subclass: TypeAlias = NicoPPAttr
+    data_split_seed: ClassVar[int] = 666  # this is the seed from the paper
```
MylesBartlett, line 113: Using the same seed won't lead to consistent splits but I guess 666 is as sensible a default as any, despite the numerological associations

conduit/data/datasets/vision/nico_plus_plus.py
```diff
@@ -188,7 +190,7 @@ def _extract_metadata(self):
         meta = json.load(open(self._base_dir / "dg_label_id_mapping.json", "r"))
 
         def _make_balanced_testset(
-            df: pd.DataFrame, seed: int = 666, verbose: bool = True, num_samples_val_test: int = 75
+            df: pd.DataFrame, seed: int, num_samples_val_test: int, verbose: bool = True
```
MylesBartlett, line 193:
```suggestion
            df: pd.DataFrame, *, seed: int, num_samples_val_test: int, verbose: bool = True
```

conduit/data/datasets/vision/nico_plus_plus.py
```diff
@@ -152,90 +139,102 @@ def __init__(
...
-            columns={0: "subclass", 1: "superclass", 2: "filename"}
+        if not all(
+            (self._base_dir / "public_dg_0416" / "train" / attr).exists() for attr in NicoPPAttr
+        ):
+            return False
+        if not (self._base_dir / "dg_label_id_mapping.json").exists():
+            return False
+        return True
+
+    def _extract_metadata(self):
```
MylesBartlett, line 187:
```suggestion
    def _extract_metadata(self) -> None:
```

conduit/data/datasets/vision/nico_plus_plus.py
```diff
@@ -152,90 +139,102 @@ def __init__(
...
+        attributes = ["autumn", "dim", "grass", "outdoor", "rock", "water"]  # 6 attrs, 60 labels
+        meta = json.load(open(self._base_dir / "dg_label_id_mapping.json", "r"))
+
+        def _make_balanced_testset(
+            df: pd.DataFrame, seed: int, num_samples_val_test: int, verbose: bool = True
+        ) -> pd.DataFrame:
+            # each group has a test set size of (2/3 * num_samples_val_test) and a val set size of
+            # (1/3 * num_samples_val_test); if total samples in original group < num_samples_val_test,
+            # val/test will still be split by 1:2, but no training samples remained
+            import random
```
MylesBartlett, line 198: Does this import need to be local?

conduit/data/datasets/vision/nico_plus_plus.py
```diff
@@ -152,90 +139,102 @@ def __init__(
...
+            random.seed(seed)
+            val_set, test_set = [], []
+            for g in pd.unique(df["g"]):
+                df_group = df[df["g"] == g]
+                curr_data = df_group["filename"].values
+                random.shuffle(curr_data)
+                split_size = min(len(curr_data), num_samples_val_test)
+                val_set += list(curr_data[: split_size // 3])
+                test_set += list(curr_data[split_size // 3 : split_size])
+            if verbose:
```
MylesBartlett, line 209:
```suggestion
            if verbose:
```
Could probably do away with this verbose mode (or at least the toggleability of it)

conduit/data/datasets/vision/nico_plus_plus.py
tmke8: I guess the only reason is that it's rarely needed.

conduit/data/datasets/vision/nico_plus_plus.py
```diff
@@ -152,90 +139,102 @@ def __init__(
...
+            combined_set.update(dict(zip(test_set, [2 for _ in range(len(test_set))])))
+            df["split"] = df["filename"].map(combined_set)
+            df["split"].fillna(0, inplace=True)
+            df["split"] = df.split.astype(int)
+            return df
+
+        all_data = []
+        for c, attr in enumerate(attributes):
+            for label in meta:
+                folder_path = self._base_dir / "public_dg_0416" / "train" / attr / label
```
MylesBartlett, line 222: Maybe make "public_dg_0416" a class variable?

tmke8: something like

```python
    subpath: ClassVar[Path] = Path("public_dg_0416") / "train"
```
?

MylesBartlett: Sure, if only the full subpath is ever needed (which seems to be the case)

MylesBartlett (approved):

conduit/data/datasets/vision/nico_plus_plus.py
tmke8: who would have thought that we're implementing a method from Satanists

conduit/data/datasets/vision/nico_plus_plus.py
```diff
@@ -152,90 +143,100 @@ def __init__(
...
+                curr_data = df_group["filename"].values
+                random.shuffle(curr_data)
+                split_size = min(len(curr_data), num_samples_val_test)
+                val_set += list(curr_data[: split_size // 3])
+                test_set += list(curr_data[split_size // 3 : split_size])
+            self.logger.info(f"Val: {len(val_set)}, Test: {len(test_set)}")
+            assert len(set(val_set).intersection(set(test_set))) == 0
+            combined_set = dict(zip(val_set, [NicoPPSplit.VAL.value for _ in range(len(val_set))]))
+            combined_set.update(
+                dict(zip(test_set, [NicoPPSplit.TEST.value for _ in range(len(test_set))]))
```
tmke8, line 214: If I understand this code correctly, it is incredibly inefficient, but whatever.

PR #313, opened 2023-05-30
Fix all pyright errors

tmke8:


tmke8:
I ran the slow tests as well.

PR #317, opened 2023-06-07
Turn `_sample_x()` into valid overrides

tmke8:


tmke8:
This is only type annotations changes; no runtime changes, so I'll just merge this.

PR #386, opened 2023-09-28
Don't use settable properties in protocols

tmke8:
I'm not actually sure why it doesn't work, but I tried on both Python
3.9 and 3.10 and it caused the same problem in both.

But I think specifying it as an attribute results in the same effect
anyway, so we'll just do that.

### DIFF
--- a/conduit/data/datasets/utils.py
+++ b/conduit/data/datasets/utils.py
@@ -696,7 +696,7 @@ def stratified_split(
                     f"No samples belonging to superclass 'y={superclass}' exist in the dataset of "
                     f"type {dataset.__class__.__name__}."
                 )
-            if isinstance(value, float):
+            if isinstance(value, (float, int)):
                 if not 0 <= value <= 1:
                     raise ValueError(
                         "All splitting proportions speicfied in 'train_props' must be in the "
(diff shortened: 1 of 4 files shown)


conduit/data/datasets/utils.py
```diff
@@ -696,7 +696,7 @@ def stratified_split(
                     f"No samples belonging to superclass 'y={superclass}' exist in the dataset of "
                     f"type {dataset.__class__.__name__}."
                 )
-            if isinstance(value, float):
+            if isinstance(value, (float, int)):
```
tmke8, line 699: Fixing an unrelated problem that pyright is suddenly complaining about.

tests/metrics_test.py
```diff
@@ -19,9 +19,9 @@ def generator() -> torch.Generator:
 
 
 def assert_close(a: Tensor | np.ndarray | float, b: Tensor | np.ndarray | float, /) -> None:
-    if isinstance(a, (np.ndarray, float)):
+    if isinstance(a, (np.ndarray, float, int)):
         a = torch.as_tensor(a, dtype=torch.float32)
-    if isinstance(b, (np.ndarray, float)):
+    if isinstance(b, (np.ndarray, float, int)):
```
tmke8, line 24: Fixing an unrelated problem that pyright is suddenly complaining about.

conduit/data/datasets/utils.py
MylesBartlett: how capricious

olliethomas (approved):

MylesBartlett (approved):

PR #468, opened 2024-02-09
Add new methods for transforming tabular data

tmke8:
You can now do it like this:

```python
    tf = ZScoreNormalize()
    train.fit_transform_(tf)
    test.transform_(tf)
```
In the next step, I could add this code to `_get_splits()` in the tabular data module.

I also renamed `cont_features` to `non_ohe_features` because that's the thing we actually care about. And then I removed `disc_features` because the same information is contained in `feature_groups`.

### DIFF
--- a/conduit/data/datasets/tabular/base.py
+++ b/conduit/data/datasets/tabular/base.py
@@ -7,12 +7,22 @@ from torch.functional import Tensor
 
 from conduit.data.datasets.base import CdtDataset, I, S, Y
 from conduit.data.structures import TargetData
-from conduit.transforms.tabular import TabularTransform
+from conduit.transforms.tabular import TabularNormalize
 
 __all__ = ["CdtTabularDataset"]
 
 
 class CdtTabularDataset(CdtDataset[I, Tensor, Y, S]):
+    """A dataset for tabular data.
+
+    :param x: The input features.
+    :param y: The target data.
+    :param s: The sensitive attribute data.
+    :param non_ohe_indexes: The indexes of the non-one-hot-encoded features. If None, all features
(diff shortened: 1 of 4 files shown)


conduit/data/datasets/tabular/base.py
```diff
@@ -21,37 +32,28 @@ def __init__(
...
+        return self.x[:, self.non_ohe_indexes]
 
-    @property
-    def x_disc(self) -> Optional[Tensor]:
-        if self.disc_indexes is None:
-            return None
-        return self.x[:, self.disc_indexes]
+    def fit_transform_(self, transform: TabularNormalize) -> None:
+        """Fit a transformation to the non-one-hot-encoded features and transfrom in-place."""
+        if (x_non_ohe := self.x_non_ohe) is not None:
```
MylesBartlett, line 53: Use `some` from ranzen?

conduit/data/datasets/tabular/base.py
```diff
@@ -21,37 +32,28 @@ def __init__(
...
+            self.x[:, self.non_ohe_indexes] = transform.fit_transform(x_non_ohe)
+
+    def transform_(self, transform: TabularNormalize) -> None:
+        """Transform the non-one-hot-encoded features in-place."""
+        if (x_non_ohe := self.x_non_ohe) is not None:
```
MylesBartlett, line 58: Same applies here.

MylesBartlett (approved):

conduit/data/datasets/tabular/base.py
tmke8: :+1: 

(I also just spotted a typo in the docstring.)

PR #469, opened 2024-02-11
Add ACS data module which applies tabular transformations

tmke8:
I added a new class `CdtTabularDataModule` which applies the transformations in `_setup()`. Not completely sure whether that's the best place to do it.


conduit/data/datamodules/tabular/base.py
```diff
@@ -0,0 +1,43 @@
...
+
+    @override
+    def _setup(self, stage: Stage | None = None) -> None:
+        super()._setup(stage)
+
+        if some(tf := self._default_transforms()):
+            assert not isinstance(
+                self._train_data, InstanceWeightedDataset
+            ), "Combination of instance weights and tfs is not yet supported for tabular data."
+            assert self._train_data is not None
```
MylesBartlett, line 37:
```suggestion
            assert some(self._train_data)
```

conduit/fair/data/datamodules/tabular/acs.py
```diff
@@ -0,0 +1,43 @@
...
+    @override
+    def _get_splits(self) -> TrainValTestSplit[ACSDataset]:
+        data = ACSDataset(
+            setting=self.setting,
+            survey_year=self.survey_year,
+            horizon=self.horizon,
+            survey=self.survey,
+            states=self.states,
+        )
+        val_data, test_data, train_data = data.random_split(
```
MylesBartlett, line 36: Didn't we want to split conditionally?

tests/data_test.py
```diff
@@ -462,3 +463,23 @@ def test_tabular_transform() -> None:
...
+    assert acs_income.train_data.feature_groups is not None
+    assert acs_income.train_data.feature_groups[0] == slice(2, 10)
+    assert acs_income.train_data.feature_groups[1] == slice(10, 33)
+    assert acs_income.train_data.x.shape == (13_360, 729)
+    assert acs_income.train_data.s.shape == (13_360,)
+    assert acs_income.train_data.y.shape == (13_360,)
+    assert acs_income.train_data.non_ohe_indexes == [0, 1]
+
+    # Verify that the non-one-hot-encoded features are normalized.
+    assert acs_income.train_data.x[:, 0].mean() == pytest.approx(0.0, abs=1e-5)
```
MylesBartlett, line 481: Could also check for unit variance I suppose but I think the test is sufficiently rigorous as it is

MylesBartlett (approved):

conduit/fair/data/datamodules/tabular/acs.py
tmke8: Hmm, you mean that hyaline and support-matching will want to do their own splitting, based on W&B artifacts? If so, then I agree it's a problem.

tmke8: You probably meant something else by "split conditionally" but I don't know what.

tmke8:
It might make sense to move the splitting into the Dataset. Then you'd call it with `ACSDataset(split=Split.TRAIN)` and because the splitting is deterministic, this would work even when you independently call `ACSDataset(split=Split.TEST)`.

Hmm, though the sizes of the splits would then also need to be parameters.

PR #493, opened 2024-02-27
Don't include the group in the features in ACS

tmke8:


conduit/fair/data/datasets/acs.py
```diff
@@ -123,15 +137,59 @@ def _employment_filter(data: pd.DataFrame) -> pd.DataFrame:
...
+    features=[
+        'AGEP',  # age; for range of values of features please check Appendix B.4 of Retiring Adult:
+        # New Datasets for Fair Machine Learning NeurIPS 2021 paper
+        'COW',
+        'SCHL',  # educational attainment
+        'MAR',  # marital status
+        'OCCP',
+        'POBP',
+        'RELP',
+        'WKHP',
```
tmke8, line 179: If anyone knows what these other features mean, please speak up.

conduit/fair/data/datasets/acs.py
```diff
@@ -180,6 +238,9 @@ def __init__(
             features, label, group = dataset.df_to_numpy(acs_data)
             non_ohe_indexes, feature_groups = None, None
 
+        # Make sure `group` is zero-indexed.
+        group -= group.min()
```
MylesBartlett, line 242: Using the inverse unique indices would robustify this as we often desire the labels to be contiguous as well as 0-indexed; perhaps the contiguity is guaranteed at some prior point, however

conduit/fair/data/datasets/acs.py
MylesBartlett: Features are features; we don't care about their underlying meaning, only their value and the distributions they follow

conduit/fair/data/datasets/acs.py
```diff
@@ -100,14 +100,28 @@ def _employment_filter(data: pd.DataFrame) -> pd.DataFrame:
...
 
+def _adult_filter(data: pd.DataFrame) -> pd.DataFrame:
+    """Mimic the filters in place for Adult data.
+
+    Adult documentation notes: Extraction was done by Barry Becker from
+    the 1994 Census database. A set of reasonably clean records was extracted
+    using the following conditions: ((AAGE>16) && (AGI>100) && (AFNLWGT>1)&& (HRSWK>0))
+    """
+    df = data
+    df = df[df['AGEP'] > 16]
```
MylesBartlett, line 111: It would be much more efficient to assemble a single mask and index the dataframe once with it; indexing a pandas dataframe is not cheap!

MylesBartlett (approved):

conduit/fair/data/datasets/acs.py
tmke8: Hmm, that's not a bad idea, though then I should print out the mapping somewhere so that the experimenter can reconstruct the original values

